PR #7, opened 2018-12-20
 Added GCC OpenMP 4.0 (GOMP 4.0) dependency for Aer. 

delapuente:
This does not seem to be a problem for modern building environments but a recommendation for Qiskit Aer is to check its dependencies before trying to load its components, providing the user with information about the several problems that could arise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 cache: pip
 sudo: false
+addons:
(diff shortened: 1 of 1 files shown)


diego-plan9 (approved):
Great troubleshooting session! Thanks @delapuente

PR #16, opened 2018-12-26
Add minimal "make doc" sphinx project

diego-plan9:
Add the structure of a basic `make doc` command that can be invoked from the top level folder and renders the documentation, mimicking terra.

Upon invoking `make doc`, the rendered docs are placed in:
```
doc/_build/html/index.html
```
(a `make clean` command is also included that deletes the rendered files).

### Details

Please note this does **not** include the visual theme yet (it uses the default Sphinx theme, `alabaster`) and is meant to provide a basic usable Sphinx installation that we can build upon. Also, the Sphinx version used is the most recent one (`Sphinx==1.8.3`) - note that in terra and other projects an older version is used (mostly because of `autodoc` and the theme): this is not a strong decision and mostly a mean to explore the most recent features, and if eventually the cost of using the latest Sphinx version is meaningful (ie. causing annoyances for people working on the documentation, incompatibilities with the autodoc, etc) reverting to the older version is always a safe option.


jaygambetta (approved):
Thanks deigo. Let’s see how the new version goes

PR #30, opened 2019-01-09
Add asv benchmarks

mtreinish:
### Summary

This commit adds an asv benchmark suite to the tests include in this
repo. This will enable us to track the performance of the overall qiskit
project over time as we make commits.


ajavadia (changes requested):
1- I don't think we should include Qasm files here. They are quite slow to load, and quite unreadable. Instead we should have python scripts that generate those circuits (like the VQE.py and State_Tomography.py performance tests in Terra). Where do these qasms come from anyway?

2- The LiH and H2 are not used anywhere so should not be included.

mtreinish:
For the qasms files sure I can remove those, but it's worth pointing out that they're not actually part of what is benchmarked they're just used to setup large circuit examples quickly without a lot of code. The only thing benchmarked is calling compile() on those circuits. For this role I think qasm works pretty well. As for the source of the files the pea_3_pi_8.qasm comes from the openqasm examples:  https://github.com/Qiskit/openqasm/blob/master/examples/generic/pea_3_pi_8.qasm and the test_eoh_qasm.qasm file I actually got from you. IIRC you said it was generated by aqua and I've been using it to test performance fixes locally. 

Yeah, I'll remove the LiH and H2 directories. I forgot to remove them when I dropped the vqe benchmark. This patch is just a port of the work I did several months ago in terra which included a vqe benchmark based on the performance test in there. (although that doesn't work anymore)

mtreinish:
This is mostly good to go now, it feels like it's a good starting point to build off of for more benchmarks and adding more backwards compat glue code so we can compare earlier releases to what we have today.

I'm still debugging one issue though before I remove the WIP tag. For a good portion of the commits when doing asv run over a range it raises either a `ModuleNotFoundError` or a `AttributeError` when trying to import something from `qiskit` (the error depends on what the import line is and what it fails to find). My assumption is that this is caused by the namespace package leaving something in place after installing and uninstalling repeatedly and something gets confused so we're left without any of the modules from terra in the qiskit namespace. But I haven't been able to confirm what the root cause is yet.

mtreinish:
This is looking good so far, it seems to avoid the install issues I was facing before. I'm running a quick benchmark on all terra commits from 0.7.0 to master to verify it works (on a quicker benchmark some of them can take a a while, like quantum volume) to verify there are no install issues. Once that passes without error I'll remove the WIP tag.

ajavadia (dismissed):

ajavadia (approved):

PR #31, opened 2019-01-10
Add "autodoc" target for Sphinx autogeneration

diego-plan9:
### Summary

Add a `make autodoc` target for generating the documentation for the
`qiskit` modules. 

At the moment it assumes the packages are installed via pip and available in the PYTHONPATH, aiming at the autogeneration from stable versions in the CI - if not, the generation of the documentation proceeds without the automatically generated content. 

### Details and comments

Next steps would be including `aqua` autogenerated documentation (currently it only looks for the `qiskit/*` modules), and on lower priority add some convenience for user to generate it without installing the modules (at the moment it can be done by manually invoking `make doc` with the right PYTHONPATH and some degree of manual effort).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,7 +114,7 @@ test/python/test_qasm_python_simulator.pdf
 
 doc/_build/*
 
-doc/**/_autodoc
+doc/**/autodoc
 
 qiskit/terra/bin/*
 
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ endif
 
 doc: autodoc
 	make -C doc html
+	rm -rf doc/_build/html/_static/font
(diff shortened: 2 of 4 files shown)


diego-plan9:
This also changes the generated style a bit, as the `better-apidoc` (that allowed customizing the template a bit) has been dropped - some effort was needed for updating it to the newer Sphinx version, and it seemed it was not providing such a high value and making the flow a tad more cumbersome. The documentation structure looks a bit closer to current aqua's now (but split in several pages).

diego-plan9:
Additionally, f7d6bf4 adds the removal of some .css and font files that are not in used by our theme - chopping of ~50MB out of the generated docs (hopefully contributing to @abdonrd 's happiness by a small amount).

jaygambetta (approved):

PR #32, opened 2019-01-10
Forces reinstallation of qiskit-terra

### LINKED ISSUE #27: Qiskit is failing when updating to 0.7 from previous versions
The root cause of https://github.com/Qiskit/qiskit-terra/issues/1599 is as follows:

When upgrading, `pip` resolves the dependency tree of a package so that dependencies are installed before dependants. Finally, it installs the target package. During any of these installations, if there is a previous version of the package, `pip` first uninstalls the previous version, then it installs the new one.

When uninstalling a package, `pip` reads a record file listing all the files belonging to that package. Previous `qiskit` included all Terra files (now belonging to the `qiskit-terra`) and those get removed from the file system. Indeed, `pip` does not check if the files being removed belong to other packages.

delapuente:
Fix #27

The proposed solution force reinstalling `qiskit-terra` after `pip`
finishes a regular installation. The new installation avoids installing
`qiskit-terra` dependencies again and should take advantage of the `pip`
cache to avoid downloading twice the same Terra version.


jaygambetta:
@diego-plan9 and @mtreinish can you look at this.

diego-plan9 (approved):
This should work, thanks @delapuente ! As mentioned in the description and in the detailed comment, hopefully this will solve the upgrades from <0.7. Manoel also offered some insight, which I think we have covered by https://github.com/Qiskit/qiskit/blob/master/docs/install.rst#troubleshooting.

jaygambetta:
But should we release a new pip

PR #36, opened 2019-01-13
Update tooling to "docs", add terra to index

diego-plan9:
### Summary

Update the tooling so `docs/` is used instead of `doc/`, and add the Terra index document to the main toctree (along with the roadmap), tweaking some things in the process (for example, removing the changelog as it is not present, and similar).

The goal of the PR is to have `make doc` work after #35, so further changes (structure, etc) can be done and tested more easily.


jaygambetta:
This is great. You read my mind before I made the issue.

jaygambetta (approved):

PR #42, opened 2019-01-15
Automate building of qiskit-tutorials to make doc

mtreinish:
### Summary

This commit changes the statically generated markdown from the
qiskit-tutorials repo to use automatically generated output on each run
of make doc. It will clone the qiskit-tutorials copy the relevant
notebooks to the proper place in the docs tree and then leverage the
nbsphinx extension to render those as part of the sphinx build. This
means we'll automatically get updates to the tutorials in the docs as we
start to use them for more guides in terra and we don't have to manually
update them.

### Details and comments

This depends on having nbsphinx and pandoc installed to run.


jaygambetta:
I think we should not automate it from the notebooks. It is ok it started from them but we should let the documentation evolve and the tutorials to be examples to show how to use it.

PR #81, opened 2019-02-12
Make proofreading changes in roadmap

Travis-S:
### Summary

This PR makes some proofreading changes in the roadmap, mostly of a typographical and grammatical nature.


jaygambetta:
These are great. 

@ajavadia and @diego-plan9 one question i have is are we using the convention of r0.6 or v0.6 we have been sticking with r from the past i dont mind but we need to keep this constant. So could you comment here what you want and then we can either change it from r to v in the pr or get @Travis-S to not make this change.

ajavadia:
I don't mind either way. I agree we should be consistent. I did a quick search over our github repositories and there are *very few* places we actually refer to versions. When we do, we use v.

https://github.com/Qiskit/qiskit/blob/7524ddd0a80c061c5bb31ca617c142ff33a6f861/docs/aqua/chemistry/release_history.rst

https://github.com/Qiskit/qiskit-tutorials/blob/dccb9b8044679eb0bdb8389c94355a30a8162fc3/releases/Terra_0.6.ipynb

pistoia (approved):

pistoia:
I approved this PR because Travis used the "v" convention, which is consistent with Ali's findings.

diego-plan9:
I'd actually argue for not having `v` or `r` - just the version number, as I feel it is somehow implied that we are talking about a release/version. But either way sounds fine to me too as long as we are consistent! :+1:

PR #87, opened 2019-02-13
[WIP] Added lang directory to docs

rraymondhp:
### Summary
I moved the contents of non english docs at `qiskit-terra` to `qiskit/docs`.  

### Details and comments
This is an action to deal with the issue below:
https://github.com/Qiskit/qiskit-terra/issues/1677


jaygambetta (commented):
I would split this into two Japan and German and get someone that translates this to do it.

diego-plan9:
I'm labeling it as `on hold`, as I think there is a pending conversation with @lzdanski about translating via translation strings, as opposed to including the full files directly.

jaygambetta:
sounds good. But @diego-plan9 you do you agree that the readme translations of terra do not need to be in this pr.

rraymondhp:
@jaygambetta  I agree that the terra Readme should not be in here. I will be working to remove them. 

@diego-plan9  Thanks for labeling it as on-hold. Will continue editing and let you know if it is done

diego-plan9:
> sounds good. But @diego-plan9 you do you agree that the readme translations of terra do not need to be in this pr.

Yes, this repository contains the sphinx documentation - I also think that individual `README.md` for each component belong to each specific repository.

> @diego-plan9 Thanks for labeling it as on-hold. Will continue editing and let you know if it is done

Thanks - depending on the conversations with Laura and Paul, the actual format for the translations might not be the one used in this PR (instead of plain .rst files, we might switch to something easier to mantain), but progress on the actual translation one way or the other is more than welcome! We can adjust later the technical aspects if needed :+1:

lzdanski:
**Update**: We're evaluating some potential tooling options to help make translation/localization easier. Until then, please continue to add translations to the PR as @diego-plan9 described.

derivation:
Closing this PR since it's being addressed in #475

Issue #80, opened 2019-02-12
Publish both stable and development documentation


diego-plan9:
### What is the expected behavior?

From #39:

> allow keeping both the "stable" documentation (the one relevant to published packages) and the "master" (the one for more bleeding edge users) available (and maybe older versions too?).

jaygambetta:
Are we going to do this with just a section for development documentation in each element? as a subsection or a new folder that is development form the top level.

diego-plan9:
My thought was along the lines of having two separate URLs (for example, `/documentation` and `documentation/0.7/`), each one containing the fully rendered .html files. This way we can "tie" their deployment to meta-qiskit releases and treat each of the two as separate and isolated entities, moving them around as needed.

jaygambetta:
Ok

diego-plan9:
Labelling also as `theme`, as we need to find a UI-sensible way of linking to different versions or toggling between them.

jaygambetta:
rather than doing it with a 0.7 can we have something like 

<img width="429" alt="image" src="https://user-images.githubusercontent.com/15731369/52907747-b69e6480-3235-11e9-8850-c44c58ae3b28.png">

where you can toggle between the stable and master version of the doc. Or it could be dev/0.7/0.6/0.5 etc

diego-plan9:
But would that imply that the only part of the documentation that has two versions is the "API references"?  It is likely to be the section that contains more changes, but I believe the rest of the documentation might change between versions as well - that is partly why I suggested just having each version self-contained in different URLs (ie. as two independent sets of .html files, etc).

jaygambetta:
I am fine to have it at the top I just dont think the 5 files should not change that much. But I agree if contributing and installing changed it should go up that top like 

<img width="316" alt="image" src="https://user-images.githubusercontent.com/15731369/63217250-d04f3480-c110-11e9-8881-bebadbf7c798.png">

jaygambetta:
closing and starting a new issue #466 as we have changed the docs

PR #89, opened 2019-02-13
moved Aer tutorial images

derivation:
### Summary
Edited /Makefile to point `make autodoc_aqua` and `make autodoc_chemistry` point to the correct directories.

Moved images and fixed their references in Aer tutorials.


Makefile
```diff
@@ -21,13 +21,13 @@ endif
 autodoc_aqua:
 ifneq ($(PATH_AQUA), )
 	sphinx-apidoc --output docs/autodoc --separate --implicit-namespaces --module-first -d 16 \
-		$(PATH_QISKIT)/qiskit_aqua
+		$(PATH_AQUA)/qiskit/aqua
```
diego-plan9: @derivation Seems we double-posted - I included some related changes in #90 ! Actually note that the current version of the `Makefile` is meant to render the "stable" version of the autodoc (as in, it expects that the packages installed are the ones from `pip install qiskit-foo`), hence looking for `qiskit_aqua` and `qiskit_chemistry`.

The support for the "development" version of the autodocs involves some extra work (to be implemented during #80), and in particular involves handling gracefully the new aqua and chemistry modules (`qiskit.aqua` instead of `qiskit_aqua`). Since some extra tuning is required for supporting both versions of the documentation, and considering the Makefile is mostly used by the CI for creating the documentation and pushing it, would you mind keeping this changes local for the time being and not including them in this PR?

derivation: Ah! My mistake. I'll retract this change.

jaygambetta (approved):

PR #115, opened 2019-02-18
Adding a community extension draft

jaygambetta:
This is a proposal for the community extensions page. 

Looks like 

image

### DIFF
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -1,7 +1,7 @@
 Qiskit Community Extensions
 ===========================
 
-Qiskit has been designed to modular in nature. It is extendable in many different ways and 
+Qiskit has been designed with modularity in mind. It is extensible in many different ways and 
 it is important to us to highlight the community projects
 
 Providers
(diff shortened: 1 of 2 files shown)


atilag (dismissed):

docs/community.rst
```diff
@@ -0,0 +1,41 @@
+Qiskit Community Extensions
+===========================
+
+Qiskit has been designed to modular in nature. It is extendable in many different ways and
```
atilag: ... designed **with modularity in mind**. It's extendable...  ?

atilag (dismissed):

docs/community.rst
derivation: +1 on "...designed with modularity in mind."

Keep "It is" to stay formal.

"extensible" is more common in this use case than "extendible".

derivation (dismissed):
I approve. My in-line comments are just suggestions.

jaygambetta:
could you make the changes they are good.

nonhermitian:
I think we should explicitly make clear the qiskit versions that the extensions work for.

derivation (approved):

jaygambetta:
@nonhermitian that is a good idea but that should go in the bullets like 

**qiskit version** 0.7

derivation:
@nonhermitian good idea re: bullets of Qiskit versions. Noted for when projects start getting added to the list.

PR #118, opened 2019-02-19
Fix/101 documentation header

### LINKED ISSUE #101: the documenation header and the qiskit.org header should look the same.
The qiskit is too far to the left and the documentation needs to be added to the documentation

carlosazaustre:
### Summary
Styles fixed in order to the header in both sites (Qiskit home and Documentation page) looks the same.

### Details and comments
Fix #101


diego-plan9 (approved):
Thanks @carlosazaustre !

PR #119, opened 2019-02-19
Issue 78 fix meta for latest pip

### LINKED ISSUE #78: IndexError occurs when installing qiskit 0.7.2 with pip 19.0.2
### Informations

- **Qiskit version**: 0.7.2
- **Python version**: 3.6.7
- **Operating system**: Ubuntu 18.10, macOS High Sierra
- **pip version**: 19.0.2

`IndexError` occurs when installing qiskit 0.7.2 with pip 19.0.2. pip <= 19.0.1 does not have this issue.

Run the following commands.
```
$ python -m venv test
$ source test/bin/activate
$ pip install -U pip wheel setuptools
$ pip install qiskit
```

Get the following result.
```
Building wheels for collected packages: qiskit
  Building wheel for qiskit (setup.py) ... done
Exception:
Traceback (most recent call last):
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/cli/base_command.py", line 179, in main
    status = self.run(options, args)
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/commands/install.py", line 355, in run
    session=session, autobuilding=True
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/wheel.py", line 980, in build
    python_tag=python_tag,
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/wheel.py", line 813, in _build_one
    python_tag=python_tag)
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/wheel.py", line 821, in _build_one_inside_env
    wheel_path = builder(req, temp_dir.path, python_tag=python_tag)
  File "/home/ima/envs/test/lib/python3.6/site-packages/pip/_internal/wheel.py", line 898, in _build_one_legacy
    return os.path.join(tempd, sorted(os.listdir(tempd))[0])
IndexError: list index out of range
```

Update(Feb 13): I confirmed that the same errors occurs on mac as well.

Update(Feb 14): I confirmed that pip==19.0.2 causes the error; pip<=19.0.1 has no problem.
https://github.com/pypa/pip/commit/61fd802b71edd61ed96ccf748b7f126eb72e6caa seems cause it.

Install without error.

### Suggested solutions

Use pip==19.0.1 for a while.  I recommend the following workaround to create a new environment for a while.
```
$ python -m venv test
$ source test/bin/activate
$ pip install -U pip==19.0.1 wheel setuptools
$ pip install qiskit
```

delapuente:
Fix #78 

Forces the process to finish with an error (which is semantically more accurate) making `pip` fallback to installing from sources.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ except:
 
 setup(
     name="qiskit",
-    version="0.7.2",
+    version="0.7.3",
     description="Software for developing quantum computing programs",
     long_description="Qiskit is a software development kit for writing "
                      "quantum computing experiments, programs, and "
--- a/setup.py
+++ b/setup.py
@@ -54,8 +54,8 @@ try:
         creation."""
 
         def run(self):
-            """Do nothing so the command intentionally fails."""
-            pass
+            """Intentionally terminating the process with an error code."""
+            sys.exit(-1)
 
 
     _COMMANDS['bdist_wheel'] = _bdist_wheel


mtreinish (approved):
LGTM, I tested this locally and the `qiskit` package is installable (albeit with an ugly error displayed by pip) with the latest version of pip, 19.0.2. Without this patch it fails as reported in #78. 

I do hope at some point we can remove this workaround to reinstall terra, because it makes debugging install issues much trickier.

derivation (approved):

derivation:
@mtreinish do you think it worth while to add an Issue calling for a more direct solution in your own words?

mtreinish:
@derivation I'm not sure there is an alternative, the need for the workaround is caused by the qiskit python package owning the qiskit python namespace before the 0.7, which isn't a condition that will ever change. If we want to make sure there is a seamless upgrade process from anyone using qiskit<0.7 to a newer version we'll probably always need that. My comment before was just my idle thoughts on how I dislike that it's needed at all.

derivation:
@mtreinish now I understand :)

PR #122, opened 2019-02-19
Removing installation from under aer

### LINKED ISSUE #84: Removing installation from under aer
Can we make the installation under aer go away and have all the content in the global install section?

derivation:
Fixes #84 

Requires follow-up called for in https://github.com/Qiskit/qiskit-aer/issues/59


jaygambetta (dismissed):
This looks. I still think we don't need an overview of getting started. I think these what you have in the getting started should just be what is shown when I click on qiskit aer. Does this make sense?

derivation:
Yes, I understand. Have the ToC do the work, instead of a landing page for Getting Started.

jaygambetta:
but in your nice layout :-)

jaygambetta (approved):

PR #124, opened 2019-02-19
styles for topics

### LINKED ISSUE #51: The .. topic:: directive does not get rendered properly
Based on the template used to render the documentation's RST files in HTML, the `.. topic::` directive does not render properly.  Consider, for example, the following snippet of RST code:
```
.. topic:: Extending the Algorithm Library

    Algorithms and many of the components they use have been designed to be
    pluggable. A new algorithm may be developed according to the specific Application Programming
    Interface (API) provided by Aqua, and by simply adding its code to the collection of existing
    algorithms, that new algorithm  will be immediately recognized via dynamic lookup,
    and made available for use within the framework of Aqua.
    Specifically, to develop and deploy any new algorithm, the new algorithm class should derive from
    the ``QuantumAlgorithm`` class.  Along with any supporting  module, for immediate dynamic
    discovery, the new algorithm class can simply be placed in an appropriate folder in the
    ``qiskit_aqua\algorithms`` directory, just like the existing algorithms.  Aqua also allows for
    :ref:`aqua-dynamically-discovered-components`: new components can register themselves
    as Aqua extensions and be dynamically discovered at run time independent of their
    location in the file system.  This is done in order to encourage researchers and
    developers interested in :ref:`aqua-extending` to extend the Aqua framework with their novel
    research contributions.
```
Using a regular template, this is how the `.. topic::` directive above gets rendered:
image

However, using the current template, the `.. topic::` directive above does not get rendered in any special way and this is what one gets:
image

Can the current template get augmented to treat the `.. topic::` directive as all the other directives?  Alternatively, can we use a different template?  The Aqua documentation makes extensive use of this directive.

carlosazaustre:
### Summary
Added styles to format the topic directive markup.
See an example below:
image

### Details and comments
Resolves #51


diego-plan9:
Thanks! I think the new style looks nice - pinging @pistoia for confirmation.

jaygambetta (approved):

PR #131, opened 2019-02-21
Add instructions on how to create a provider

### LINKED ISSUE #123: Add instructions on how to create a provider
Based on https://github.com/Qiskit/qiskit-tutorials/blob/master/qiskit/terra/creating_a_provider.ipynb

derivation:
fixes #123


yaelbh:
The Hadamard results printed at the end are not correct. As in the create_a_provider tutorial, they have to be identical to the Aer results two rows below. I don't understand how this happened - I went over the code and it looks identical to the code in create_a_provider. Do you have an idea what can be the cause? If not then I'll try to take another look.

yaelbh:
In the description of the Job class, `result` and `status` are listed as properties, although they are functions. In addition, the statement that `HadamardJob` performs `cancel` and `submit` may be confusing; it has empty implementations of these functions.

yaelbh:
Please be aware that the code works with Terra 0.7 and Aer 0.1 (and not necessarily with the master branches of Aer and Terra).

derivation:
Thank you for your careful review @yaelbh  (:

> Please be aware that the code works with Terra 0.7 and Aer 0.1 (and not necessarily with the master branches of Aer and Terra). 

I tested the code with Terra 0.8.0 and Aer 0.1.1, and it still works. Are you anticipating it no longer working soon based on a dev roadmap?

> The Hadamard results printed at the end are not correct.

I introduced a bug when I removed a few lines of code. Had the effect of appending qubits being measured with this line of code:

`list_of_qubits.append(operation.qubits[0])`

Fixed by ensuring only Hadamard operations get picked out as passed arguments above.

> In the description of the Job class, `result` and `status` are listed as properties, although they are functions. In addition, the statement that `HadamardJob` performs `cancel` and `submit` may be confusing; it has empty implementations of these functions.

Fixed the above.

docs/contributing_to_qiskit.rst
```diff
@@ -49,3 +49,338 @@ The Python API reference documentation is automatically generated from comments
...
+-  Return a ``BaseJob`` object that contains the results of your
+   simulator
+
+The ``run()`` method implements these operations.
+
+
+
+**Implementing the Backend Class**
+
+.. code:: ipython3
```
diego-plan9: Can you replace these blocks with:
```
.. code:: python
```
as they are not rendered currently otherwise (more context  in #133)?

derivation:
@diego-plan9 could you approve / merge this?

diego-plan9 (approved):
It's looking good, thanks @derivation and @yaelbh !

yaelbh:
> I tested the code with Terra 0.8.0 and Aer 0.1.1, and it still works. Are you anticipating it no longer working soon based on a dev roadmap?

No, but it has to be verified with every new version.

PR #145, opened 2019-02-25
fixed ident styles on TOC items

### LINKED ISSUE #67: Misleading indentation of some TOC entries
Some items/levels in the TOC are shown with the wrong indentation, hindering readability - for example:

toc

In the screenshot, the `Terra organization` entry is visually at the same level as `Overview`, but the real hierarchy is:
```
Qiskit Terra
    Overview
        Terra Organization
    The Quantum Circuit
    ...
```

Pinging @Borjagodoy - can you double check the relevant parts of the theme for the TOC? Please note that there might be other oddities potentially related, such as the indentation for an "entry that has subentries" vs the intentation for an "entry that has no subentries".

Also, it seems [the original theme](https://myyasuda.github.io/sphinx_materialdesign_theme/examples/index.html) already had the collapsing/expanding functionality baked in by default (on the sample site, with some circles around the arrows and no extra changes) - can you evaluate if we can reuse it directly, instead of tuning the html and CSS?

carlosazaustre:
### Summary
This PR resolves through styles the misleading ident on TOC. it doesn't resolve the "reloading" issue but it resolves the visual indentation.

### Details and comments
Resolves #67

### DIFF
--- a/docs/theme/static/css/theme.css
+++ b/docs/theme/static/css/theme.css
@@ -409,7 +409,8 @@ cursor: pointer;
 .sublist .toctree-l2.link-wrapper .title {
     margin: 0 16px;
 }
-li.toctree-l2 .item.title .link-wrapper {
+
+/*li.toctree-l2 .item.title .link-wrapper {
     padding-left: 16px;
 }
 .sublist .toctree-l3 .title {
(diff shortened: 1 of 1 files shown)


derivation:
Looks great to me! Excellent work (:

derivation (approved):

PR #146, opened 2019-02-25
Pin versions and document versioning policy

### LINKED ISSUE #141: Mega and procedure for update of patches
### What is the expected behaviour?

I think that the mega patches need to link with an exact qiskit element version. For instance the line 

https://github.com/Qiskit/qiskit/blob/2fdba884b2decaa38fc38315e04c1500e54e6218/setup.py#L17

should be 

```
requirements = [
   qiskit-terra==0.7.0”,
   "qiskit-aer==0.1.1,"
]
```
This way for each patch version of qiskit we know what elements are in it and we say this in the release tag. Then our rule should be if we patch an element we need to also patch the mega. We can also patch the mega with no update to the element if the mega needs a patch.

For a minor update on any element, the mega minor should be updated. So if we go to qiskit_aer 0.2.0 then the mega updates even if Terra does not. This is also true if we add a new element. If we add ignis then the mega should increase. 

Obviously, we should try to coordinate and then we can release patches and updates at the same time.

mtreinish:
### Summary

This commit pins the versions of the tracked elements to ensure that we
only ever install a single version of an element when we install the
meta-package. Additionally, to make it clear why we do this, and how we
handle versioning this also adds a doc on the procedure for handling
versioning.

### Details and comments

Fixes #141

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -42,16 +42,8 @@ The local html version of the documentation can be found at `docs/_build/html/in
 The meta-package started with version `0.7` to respect the continuity of `qiskit` versions. The
 Terra element did the same to provide continuity with its own history.
 
-Considering [`semver`](https://semver.org/), the Qiskit meta-package pins the _minor_ version
-number of each Qiskit element to get new patches automatically.
-
-Nevertheless, upon changes _minor_ or _major_ version numbers of the elements, the meta-package
-version must be updated according to the following rules:
-
-1. If a dependency increased the **major** number, increase the **major** number of the meta-package.
-2. Else:
-   1. If a dependency increased the **minor** number, increase the **minor** number of the meta-package.
-
+Details on the versioning procedure are documented here:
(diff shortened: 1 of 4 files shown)


jaygambetta (commented):
i would also just link the contributing.md file for this section to this doc. 

https://github.com/Qiskit/qiskit/blob/master/.github/CONTRIBUTING.md#versioning

jaygambetta (dismissed):

jaygambetta:
do you want this in the TOC. If so we need to add it to the index.rst file as well.

derivation:
To me,

- roadmap.rst
- components.rst
- **versioning.rst**

could all fall under a top level heading "Our Strategy" or "Development Strategy".

Your thoughts?

mtreinish:
I'll push a patch to add it to the TOC now.

@derivation I think we can make a new top level heading for those, we can do that in a followup PR. I kinda like "Development Strategy" more than "Our Strategy" but either works for me.

derivation (approved):

PR #159, opened 2019-02-27
Add constraints to pin pylint to known working version

mtreinish:
### Summary

This commit adds a constraints file so we're always running pylint and
astroid at known working version. The 2 are requirements are very coupled
and small changes in one often requires an update in the other.
Unfortunately new versions of pylint often change or add rules which
makes running it unpinned not a good decision. To ensure a consistent
environment for running pylint this commit leverages pip constraints to
make sure we always install the same version of astroid and pylint.

### Details and comments

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,0 +1,3 @@
+astroid==2.1.0
+pylint==2.2.2
+cryptography==2.5.0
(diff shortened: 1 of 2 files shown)


mtreinish:
@derivation This patch should workaround the issues with the latest pylint. It'll keep us at fixed known working versions until we're ready to bump things.

derivation (approved):

Issue #88, opened 2019-02-13
Makefile pointing to wrong directories for aqua and chem autodoc?


derivation:
@diego-plan9, can I get a sanity check from you? 

### Information

- **Qiskit version**: latest
- **Python version**: 3.7
- **Operating system**: macOS 14.6

### What is the current behavior?

`make doc` errors out while looking for Aqua and Chemistry directories

sphinx-apidoc --output docs/autodoc --separate --implicit-namespaces --module-first -d 16 \
		<PATH_QISKIT>/qiskit_aqua
~/.../qiskit-terra/qiskit_aqua is not a directory.
make: *** [autodoc_aqua] Error 1

### Steps to reproduce the problem

1. Clone all of the repos with Qiskit doc in them.
2. Run `make doc`

### What is the expected behavior?

The Makefile should be pointing to the correct directories using `pip show` and `grep`.

### Suggested solutions

I've edited the Makefile in a personal fork to make things work, but I'm wary that there's a reason why the Makefile is the way it is today that I'm not aware of, despite it not working for me locally.

diego-plan9:
Good catch @derivation !

https://github.com/Qiskit/qiskit/blob/master/Makefile#L24
```
$(PATH_QISKIT)/qiskit_aqua
```
should indeed be
```
$(PATH_AQUA)/qiskit_aqua
```

PR #160, opened 2019-02-28
Qiskit documentation for characterization

yaelbh:
### Summary

Addressing #88

### Details and comments

This is a WIP. So far I wrote the section "Generating Coherence Circuits", but I have to write an additional section "Analyzing the Results".

@dcmckayibm I appreciate if you can:
- Already review what's done so far.
- Write about ZZ.

Thanks!


yaelbh:
I wrote the Analyzing the Results section. Still need to fill in a few placeholders and fix rst syntax, but the main text is essentially ready.

yaelbh:
Ready for review.
Note that `:math:` is not working, also in https://github.com/Qiskit/qiskit/blob/master/docs/aer/device_noise_simulation.rst.

dcmckayibm (dismissed):

yaelbh:
I fixed a few typos, please check that it's OK.
A couple of unfixed comments:
- "a single qubits data" - not sure if it's correct.
- The Latex/Python expressions towards the end of the document do not display well, at least not in GitHub. Also there is some inconsistency between the expressions, for example one expression has Latex ``\cos`` while the other one has Python ``np.cos``.

derivation:
@yaelbh I'll look into these

derivation:
@yaelbh please review my changes to make sure the last section is still technically valid.

yaelbh:
>  I took the liberty of writing the last two equations in a simpler form. Were there reasons why they were written in the prior form, or is the new form alright?

@dcmckayibm please check the equations.
`:math:` is totally not working for me, and from raw I can't figure what e.g. `a \, e^{-t/{T_2}^*} \, \cos(2\pi ft + \phi) + c` means.

derivation:
@yaelbh for now, https://quicklatex.com/ will let you render the latex code

derivation:
@yaelbh do you think users would benefit from any references to published research these tools were based on?

yaelbh:
@dcmckayibm please see https://github.com/Qiskit/qiskit/pull/160#issuecomment-469382678, thanks

dcmckayibm:
I can't think of a simple reference off the top of my head because these functions are kind of an amalgamation of different sources. We will have some community notebooks with more detail.

derivation (approved):

yaelbh:
Are the equations correct?

image

derivation:
@yaelbh I rewrote the equations in simpler forms by grouping terms and using trig identities. Are you asking about the mathematical correctness, or the fact that they're not rendering in that screenshot? The doc is rendered in HTML like so: doc

dcmckayibm:
Looks ok

PR #164, opened 2019-03-04
Add ignis to tracked elemnts

mtreinish:
### Summary

Now that there is an ignis package on pypi we can track it in the
meta-repository. Per our standard versioning this bumps the package
version to 0.8.0.

### Details and comments

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,11 @@ The format is based on `Keep a Changelog`_.
 `UNRELEASED`_
 =============
 
+Added
+-----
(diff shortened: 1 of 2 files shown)


jaygambetta:
can we also do the patch to terra

mtreinish:
There isn't a terra release with the backported patch yet. Once there is I can do a follow up patch to bump the terra requirement here too. We can hold off on actually pushing the 0.8.0 release here until we have terra 0.7.1.

derivation (approved):

PR #165, opened 2019-03-04
publish Ignis docs

derivation:
now that Ignis is on PyPi

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -145,7 +145,7 @@ where ``MY_API_TOKEN`` should be replaced with your token.
 
 .. note::
 
-  IBM Q Network members must specify more than just an API token, as explained below.
+  If you are an IBM Q Network members, you must specify more than just an API token by using the following commands.
 
 If you are a member of the IBM Q Network, you must pass an additional argument
 to ``IBMQ.save_account()``. The ``url`` argument can be found on your q-console
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -59,7 +59,7 @@ The final step required is to install Qiskit with the following command:
 
 .. code:: sh
 
-  pip install qiskit qiskit-aqua qiskit-ignis
+  pip install qiskit qiskit-aqua
 
 There are optional dependencies that are required to use all the visualization
 functions available in Qiskit. You can install Qiskit along with these optional
@@ -67,7 +67,7 @@ dependencies by running
 
 .. code:: sh
 
-  pip install qiskit[visualization] qiskit-aqua qiskit-ignis
+  pip install qiskit[visualization] qiskit-aqua
 
 
 .. _install_install_from_source_label:
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -12,33 +12,38 @@ Qiskit is tested and supported on the following 64-bit systems:
 
 Qiskit supports Python 3.5 or later.
 
-We recommend installing `Anaconda <https://www.anaconda.com/download/>`_, a cross-platform Python
-distribution for scientific computing. Jupyter Notebook, included in Anaconda, is recommended for
-interacting with the `Qiskit tutorials <https://github.com/Qiskit/qiskit-tutorial>`_.
+We recommend installing `Anaconda <https://www.anaconda.com/download/>`_, a
+cross-platform Python distribution for scientific computing. Jupyter Notebook,
+included in Anaconda, is recommended for interacting with the `Qiskit tutorials
+<https://github.com/Qiskit/qiskit-tutorial>`_.
 
-Using Qiskit on Windows requires VC++ runtime components. We recommend one of the following:
+Using Qiskit on Windows requires VC++ runtime components. We recommend one of
+the following:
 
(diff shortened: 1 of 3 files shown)


derivation:
updating global install instructions to include Ignis

docs/install.rst
```diff
@@ -12,33 +12,38 @@ Qiskit is tested and supported on the following 64-bit systems:
...
+terminal window:
 
 .. code:: sh
 
   conda create -n name_of_my_env python=3
 
-This will create a minimal environment with only Python installed in it. To put yourself inside
-this environment run:
+This will create a minimal environment with only Python installed in it. To put
+ourself inside this environment run:
```
lzdanski:
```suggestion
yourself inside this environment run:
```

docs/install.rst
```diff
@@ -54,73 +59,82 @@ The final step required is to install Qiskit with the following command:
...
 
-*	 :ref:`install-terra-source`
-*	`Qiskit Aer <https://github.com/Qiskit/qiskit-aer/blob/master/.github/CONTRIBUTING.rst>`_
-*	`Qiskit Aqua <https://github.com/Qiskit/qiskit-aqua/blob/master/.github/CONTRIBUTING.rst>`_
-*	`Qiskit Chemistry <https://github.com/Qiskit/qiskit-chemistry/blob/master/.github/CONTRIBUTING.rst>`_
-*	`Qiskit IBMQ Provider <https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/.github/CONTRIBUTING.rst>`_
-
-.. _editable mode: https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs
+* :ref:`install-terra-source`
+* `Qiskit Aer <https://github.com/Qiskit/qiskit-aer/blob/master/.github/
```
lzdanski, line 103: Is there any significance in where the line breaks occur? It's the only thing standing out to me that the number of characters in each line seem different.

docs/install.rst
```diff
@@ -54,73 +59,82 @@ The final step required is to install Qiskit with the following command:
...
-token.
+`Create a free IBM Q account <https://quantumexperience.ng.bluemix.net/qx/
+login>`_ to get an API token.
 
 After logging in, navigate to your account settings.
 
 .. image:: ./images/figures/install_my_account.png
 
-Navigate to the Advanced tab and click the Generate button in the API Token section.
+Navigate to the Advanced tab and click the Generate button in the API Token
```
lzdanski: We should have a conversation about styling UI elements to be distinct in addition to capitalizing them. It helps users know what is a command vs an element to interact with. So something like: 

Navigate to the **Advanced** tab and click the **Generate** button in the *API Token*. 

It'd have to address the whole doc set, so I don't think it's a barrier to merging this PR, and also we'd want input around what we want the style to be.

docs/install.rst
```diff
@@ -133,15 +147,16 @@ where ``MY_API_TOKEN`` should be replaced with your token.
 
   IBM Q Network members must specify more than just an API token, as explained below.
```
lzdanski:
```suggestion
  If you are an IBM Q Network members, you must specify more than just an API token by using the following commands.
```

lzdanski: This is just an accessibility vocab change. "Above/Below" use the UI design to indicate order of presentation, but if you're using a screen reader you perceive it as "before/after" because they read from beginning/top to end/bottom.

lzdanski (dismissed):
Hey @derivation - I made a couple of comments and suggestions. It looks great and the only things I found were minor (typo, accessibility).

docs/install.rst
derivation: Yes, there's a character line limit that gets checked when the docs are built. The max is currently set to 100 in the build.

docs/install.rst
derivation: Oh, good idea. I'll make these bold for now.

docs/install.rst
```diff
@@ -54,73 +59,82 @@ The final step required is to install Qiskit with the following command:
 
 .. code:: sh
 
-  pip install qiskit qiskit-aqua
+  pip install qiskit qiskit-aqua qiskit-ignis
```
mtreinish: This shouldn't be needed. Since https://github.com/Qiskit/qiskit/pull/164 added ignis to the requirements for the metapackage it will install ignis for us.

docs/install.rst
```diff
@@ -54,73 +59,82 @@ The final step required is to install Qiskit with the following command:
...
-There are optional dependencies that are required to use all the visualization functions available
-in Qiskit. You can install Qiskit along with these optional dependencies by running
+There are optional dependencies that are required to use all the visualization
+functions available in Qiskit. You can install Qiskit along with these optional
+dependencies by running
 
 .. code:: sh
 
-  pip install qiskit[visualization] qiskit-aqua
+  pip install qiskit[visualization] qiskit-aqua qiskit-ignis
```
mtreinish: Same here

mtreinish (changes requested):
LGTM, just one small thing to fix on the install doc. I think this and https://github.com/Qiskit/qiskit/pull/168 are all we need before we can push the release with ignis.

mtreinish (approved):

PR #174, opened 2019-03-10
Added Ignis to the auto-generate author file

yehuda-naveh:
### Summary
Added Ignis. Wasn't sure if to also add qiskit-tutorials


derivation:
@yehuda-naveh I think it's best to leave out the tutorials' authors. The tools and the tutorials feel like different works to me.

derivation (approved):

PR #201, opened 2019-04-19
Add depth tracking to random hex circuit benchmarks

mtreinish:
### Summary

This commit adds 2 new benchmarks that use the same random hex circuit
setup but instead of measuring the time it takes to transpile() it
tracks the depth of the output circuit. This will serve as an example
for how we can track other transpiler performance metrics over time in
terra (not just depth).

### Details and comments


1ucian0 (approved):
LGTM

ajavadia (approved):

PR #195, opened 2019-04-09
Update install instructions

lzdanski:
Remove redundant install from source
Update commands and explanation

PR #207, opened 2019-04-24
Update Install from Source instructions for Qiskit Terra

lzdanski:
### Summary
The instructions for how to Install Terra from Source were not included in the TOC and removed from the contributing.rst file in the Terra repository. See https://github.com/Qiskit/qiskit-terra/pull/1850 and https://github.com/Qiskit/qiskit/pull/195

I edited the process flow and tested the steps on my machine. Qiskit Terra installation was successful.

Topic was moved from standalone to be included in the **Contributing to Qiskit** section to remain consistent with similar projects and use cases.

### Details and comments

- Tested installation steps locally on Mac OSx (High Sierra) 
- Built docs locally to check codeblock formatting and styling
- Add redirect from Install topic to Contributing topic
- Added link to Qiskit docs on Github to TOC tree


docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+
+    conda create -y -n QiskitDevenv python=3
+    conda activate QiskitDevenv
+
+4. Install the Python requirements libraries from your ``qiskit-terra`` directory in your virtual environment.
+
+.. code:: sh
+
+    pip install -r requirements.txt
+    pip install -r requirements-dev.txt
```
mtreinish, line 169: These 2 lines can be removed and just replaced by `pip install cython` which is the only thing that needs to be manually installed before we build terra. When you run `python setup.py install` below it will install the requirements. `requirements-dev.txt` are only needed for running tests and or linting checks.

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+.. code:: sh
+
+  cd qiskit-terra
+
+3. Create a virtual environment with `Anaconda <https://www.anaconda.com/distribution/>`_.
+
+.. code:: sh
+
+    conda create -y -n QiskitDevenv python=3
+    conda activate QiskitDevenv
```
mtreinish: Using anaconda isn't a requirement for terra. It'll work fine, but it adds another layer and I'm not sure it's needed for the source installation steps. Also I don't think we need to specifically mention virtualenvs, while I normally use them the specific python environment you're building and installing into doesn't matter for the instructions.

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+.. code:: sh
+
+    pip install -r requirements.txt
+    pip install -r requirements-dev.txt
+
+5. Install the qiskit modules.
+
+.. code:: sh
+
+    python setup.py build_ext --inplace
```
mtreinish: This isn't explicitly needed. Especially with the command below. When you build from the sdist setuptools will call cython and compile the generated C code into the package when it gets installed. This is only needed if we need to build those binaries into the repo next to the cython source files, like if we were installing in dev mode or running out of the repo root.

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+
+* If you want to only install ``qiskit-terra`` onto your system.
+
+.. code:: sh
+
+    python setup.py install
```
mtreinish: This works fine, but normally I like to recommend calling `pip install .` just to let pip handle things like dependencies instead of having setuptools call pip. Functionally it doesn't make a real difference (it's mostly an artifact of having used pip for years and it didn't used to handle manually installed packages as well)

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+
+
+* To get the examples working, install and run them with the following commands.
+
+.. code:: sh
+
+    pip install -e
```
mtreinish: You need to specify a path to the source repo to install in editable mode. For example if you're in the repo root you'd use `pip install -e .`.  That being said this isn't needed to get the examples working just installing terra as above will make running the examples work. All editable mode does is enable you to change the source in place and have those reflected in the installed version of terra. (it just links the installed package to the repo instead of copying the python code to site-packages)

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+
+.. code-block:: sh
+
+    conda create -y -n QiskitDevenv python=3
+    conda activate QiskitDevenv
+
+3. Ensure that the stable version is not installed in the environment.
+
+.. code:: sh
+    pip uninstall qiskit-ignis
```
mtreinish: This shouldn't be needed, since the package manager should detect version differences and automatically upgrade the package. But I just realized we neglected to bump the version in the setup.py after pushing the tag (which is a reason I want to use a library to automate this). I just pushed up https://github.com/Qiskit/qiskit-ignis/pull/165 to address this so we can remove this step.

mtreinish (dismissed):
This looks good, I like the reorganized structure. Just some inline comments about improving the instructions for installing from source. FWIW, I often just do a one-liner `pip install git+https://github.com/Qiskit/qiskit-terra.git` or `pip install git+https://github.com/Qiskit/qiskit-ignis.git` both of which will work fine (assuming you have a c compiler installed for terra to build). But that's only useful if you want it installed from source and are not going to make changes to it.

docs/index.rst
```diff
@@ -19,6 +19,7 @@ Welcome to Qiskit's documentation!
   Contributing to Qiskit <contributing_to_qiskit>
   Community Extensions <community>
   License <license>
+  Qiskit Docs on Github <https://github.com/Qiskit/qiskit>
```
derivation: @lzdanski, I don't think the doc source link should be in the global ToC in index.rst. It's best linked in the context of Contributing To Qiskit > Writing Documentation. There's already a link there.

docs/install.rst
```diff
@@ -25,6 +25,10 @@ the following:
 * `Microsoft Visual C++ Redistributable for Visual Studio 2015 <https://
   www.microsoft.com/en-US/download/details.aspx?id=48145>`_
 
+.. note::
```
derivation, line 29: @lzdanski this seems out of place. I recommend moving it to within the Install section below.

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+
+4. Install the package in `editable mode <https://pip.pypa.io/en/stable/
+   reference/pip_install/#editable-installs>`_ from the root directory of the
+   repository. The following example shows the installation for Ignis.
+
+.. code:: sh
+
+  pip install -e qiskit-ignis
+
+Install Terra from Source
```
derivation, line 116: @lzdanski I think each element's instructions should be in their own RST file to ease maintainability. What do you think?

docs/contributing_to_qiskit.rst
```diff
@@ -52,10 +51,158 @@ The Python API reference documentation is automatically generated from comments
...
+  CONTRIBUTING.md>`_
+* `Qiskit Aqua <https://github.com/Qiskit/qiskit-aqua/blob/master/.github/
+  CONTRIBUTING.rst>`_
+* `Qiskit Chemistry <https://github.com/Qiskit/qiskit-chemistry/blob/master/
+  .github/CONTRIBUTING.rst>`_
+* `Qiskit IBMQ Provider <https://github.com/Qiskit/qiskit-ibmq-provider/blob/
+  master/.github/CONTRIBUTING.rst>`_
 
+**************************
 Creating a Custom Provider
```
derivation: While you're updating this file, please delete this entire section.

docs/contributing_to_qiskit.rst
lzdanski: I agree. While trying to make nested topics in the toctree I ran into problems--still investigating whether it's something happening on my machine or in general. Since this content doesn't appear in the Terra contributing.rst file in qiskit/qiskit-terra, I think this works as a stop-gap and we can break everything into individual files at the same time.

docs/contributing_to_qiskit.rst
lzdanski: @mtreinish, I'm still learning the intricacies of pip vs python install, so I want to make sure I understand. 

Since this specific process is to install the most recent `master` branch of the qiskit-terra repo, not to install the stable qiskit-terra pip package. Will using `pip install .` just manage installing the dependencies for the local clone of Qiskit Terra or will it pull down and install Qiskit terra from pypi?

docs/install.rst
lzdanski: I don't think that this should be moved to the "install" section since it's a pre-req instead of an actual installation step, but that section isn't super clear. I've reworked it to be more consistent -- with the windows specific requirements near the supported Operating Systems.

docs/contributing_to_qiskit.rst
mtreinish: `pip install .` tells pip to install the python package located at the current working directory (which at that point in the guide is the locally cloned copy of master) This effectively is just telling pip to run python setup.py for us. With modern pip on python 3 it has no functional difference with `python setup.py install` since pip is bundled in stdlib from cpython and setuptools and setuptools will use pip to install external things where needed.

It only really makes a difference on older versions of python (or non-cpython interpreters which we don't support anyway). But, I normally recommend this for 2 reasons: 1. just as an artifact from my experience dealing with older pip/python (or on ancient linux distros like RHEL) where calling pip and setup.py may not behave the same way 2. It's what the python packaging community documentation states as the recommend way to install from a local source tree: https://packaging.python.org/tutorials/installing-packages/#installing-from-a-local-src-tree

lzdanski: Thanks so much for the explanation!

docs/index.rst
derivation: @lzdanski this looks like the last comment to be resolved, if I understood your thumbs up correctly

derivation (approved):

docs/index.rst
lzdanski: Yes, you're right! Not sure how that one snuck through.

jaygambetta (approved):

PR #211, opened 2019-04-29
[WIP] Add release notes for Terra release 0.8

mtreinish:
### Summary

This PR adds the release notes for the Terra 0.8 release. This should
provide a guide to users explaining what the new features of the release
are and any things to keep in mind when upgrading. It's a high level guide
and should cover the important details a user will need to know when
upgrading from the previous terra version.

PR #213, opened 2019-04-29
Continue Terra 0.8 release notes

mtreinish:
### Summary

This commit is a continuation of the release notes added in #211. The
notes there weren't complete and still need to be flushed out more. This
commit starts off where that left off.

### DIFF
--- a/docs/release_history.rst
+++ b/docs/release_history.rst
@@ -16,8 +16,8 @@ This release includes several new features and many bug fixes. The major new
 features are:
 
 - Introduction of the Pulse module under ``qiskit.pulse``, which includes
-  tools for building pulse commands, scheduling them on pulse channels, and
-  visualizing them.
+  tools for building pulse commands, scheduling them on pulse channels, visualization
+  and running them on IBMQ devices.
 - Improved QuantumCircuit and Instruction classes, allowing for the
   composition of arbitrary sub-circuits into larger circuits, and also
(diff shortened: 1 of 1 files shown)


jaygambetta (dismissed):

docs/release_history.rst
```diff
@@ -26,38 +36,76 @@ In addition there is also the introduction of the following new features:
...
 Upgrade
-^^^^^^^
+-------
 
 Please note that some backwards incompatible changes have been made during this
 release. The following notes contain infomation on how to adapt to these
 changes.
 
+
 IBMQ Provider
```
jaygambetta, line 55: should we put the provider number here

mtreinish: Well I had originally included this as an upgrade note for terra just about ibmq being a separate package. Since if you install terra directly when you upgrade from 0.7.x to 0.8.0, the ibmq provider will disappear.

mtreinish: But it's a good point that we don't have any release documentation for the ibmq provider, we probably should add that too

mtreinish: Ok, I added separate ibmq release notes in: https://github.com/Qiskit/qiskit/pull/213/commits/9e50b75ae6e3b41e9cee836779cf3196b6622ced

jaygambetta (approved):

PR #185, opened 2019-03-26
Add aqua and chemistry to tracked components

mtreinish:
### Summary

For the next release of qiskit-aqua and qiskit-chemistry we're going to
include qiskit-aqua and qiskit-chemistry in the list of tracked
elements.

### Details and comments

Fixes #6

PR #227, opened 2019-05-02
Bump released package versions and set meta package version to 0.9.0

mtreinish:
### Summary

This commit bumps the version numbers for the newly released versions of
terra, aer, and ignis. It also bumps the version of the package to
reflect this change per the versioning policy.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,16 @@ The format is based on `Keep a Changelog`_.
 `UNRELEASED`_
 =============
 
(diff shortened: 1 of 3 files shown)


jaygambetta:
this should also have the aqua and chemistry and provider in the dependencies

mtreinish:
aqua and chemistry are pending here: https://github.com/Qiskit/qiskit/pull/185 I just added the provider

jaygambetta (approved):

PR #235, opened 2019-05-03
Debug sphinx core dump patch DO NOT MERGE

mtreinish:
### Summary

This patch should not be merged it harvests one of the python test jobs
to run the docs build but sets up the kernel and gdb to print the
stacktrace from any core dumps that occur during the build process.


mtreinish:
Found a solution to make the jobs build in #236

PR #236, opened 2019-05-03
Switch docs jobs to run on Xenial with Python 3.7

mtreinish:
### Summary

In the process of trying to debug the floating point exception
encountered during the docs builds in #235 docs builds were run in all
our unit test environments. The python3.7 environment which runs on
xenial successfully built the docs. So instead of trying to get to the
bottom of why there is a floating point exception during the sphinx
build in the current job setup, this commit just switches to the same
node setup as the python 3.7 unit test env. This means switching to
an Ubuntu Xenial node and using Python 3.7 to run the docs builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,10 @@ matrix:
     - python: "3.5"
       env: TOXENV=lint
     - if: branch = master AND type = push
-      python: "3.5"
+      os: linux
+      dist: xenial
+      python: "3.7"
+      sudo: true
       install:
         - pip install qiskit
         - pip install -r requirements-dev.txt


mtreinish:
The xenial 3.7 job output from #235: https://travis-ci.com/Qiskit/qiskit/jobs/197514632 
compared to the 3.5 job output: https://travis-ci.com/Qiskit/qiskit/jobs/197514630

diego-plan9 (approved):
Sounds reasonable for solving the deploying - we should however investigate a bit when the rush passes, to make sure it does not point towards something deeper, just in case.

PR #244, opened 2019-05-06
Update files for 0.10.0 release

diego-plan9:
### Summary

Update the files for preparing 0.9.1 release, which include `IBMQProvider==0.2.0`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,8 +15,18 @@ The format is based on `Keep a Changelog`_.
   - **Fixed**: for any bug fixes.
   - **Security**: in case of vulnerabilities.
 
+
 `UNRELEASED`_
 =============
+
+
+`0.9.1`_ - 2019-05-06
+=====================
+
+Changed
+-------
+
+- Updated the ``qiskit-ibmq-provider`` dependency to 0.2.0 (#227).
 
 
 `0.9.0`_ - 2019-05-02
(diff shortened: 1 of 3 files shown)


diego-plan9:
Marking as on hold while publishing the 0.2.0 provider.

mtreinish:
Hmm, per the documented version policy:
https://qiskit.org/documentation/development_strategy.html#patch-version-increases
We should bump the patch version when a tracked **element** patch version increases, and we should bump the minor version when a tracked **element** minor version increases. I would be inclined to say this should be `0.10.0` instead of `0.9.1` except that this is the provider package and not actually an element (this and maybe chemistry are the only things that aren't elements). We should write that down in the versioning doc if we're going to treat non-element packages differently than elements.

mtreinish (dismissed):
LGTM, I think you got everything

CHANGELOG.rst
```diff
@@ -15,10 +15,20 @@ The format is based on `Keep a Changelog`_.
...
 =============
 
 
+`0.10.0`_ - 2019-05-06
+======================
+
+Changed
+-------
+
+- Updated the ``qiskit-ibmq-provider`` dependency to 0.2.0 (#227).
```
mtreinish: Oh, this is actually the wrong PR number, it should be 244. 227 was thepatch that bumped aer, terra, and ignis as well as added the provider for the 0.9.0 release. But we can fix this as a follow up PR, since we want to get this release up and CI is bogged down right now.

diego-plan9: Copy-paste indeed - should be fixed at 406caac

CHANGELOG.rst
```diff
@@ -15,10 +15,20 @@ The format is based on `Keep a Changelog`_.
...
 =============
 
 
+`0.10.0`_ - 2019-05-06
+======================
+
+Changed
+-------
+
+- Updated the ``qiskit-ibmq-provider`` dependency to 0.2.0 (#244).
```
mtreinish:
```suggestion
- Updated the ``qiskit-ibmq-provider`` dependency to 0.2.1 (#244).
```

mtreinish (dismissed):
Just one small typo in the changelog (which we can fix in a follow up PR) but otherwise it LGTM

mtreinish (approved):

PR #265, opened 2019-05-10
.travis.yml: The 'sudo' tag is now deprecated in Travis CI

cclauss:
### Summary
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo: false__ in your __.travis.yml__, we recommend removing that configuration_"

### Details and comments
Removed the deprecated tag from _.travis.yml_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 cache: pip
-sudo: false
 
 addons:
   apt:


derivation (approved):
After reviewing with @SooluThomas and @diego-plan9, it seems alright to simply remove this configuration without replacing it with something else.

cclauss:
Yes.  The __sudo__ tag has become a noop because __sudo__ is ALWAYS on.

PR #276, opened 2019-05-16
Update all benchmarks to work on Terra Master

ajavadia:
### Summary
This PR updates all the benchmarks to work again with recent API changes made on Terra master. These were all things coming, but were neglected until the entire benchmark suite was broken. Once this lands we should continue getting published data from the periodic benchmark runs from the point in the git history when the breaking changes in terra landed.

### Details and comments

Co-Authored-By: Matthew Treinish <mtreinish@kortar.org>


mtreinish:
So this is actually more critical now than before because since Qiskit/qiskit-terra#2337 landed the benchmarks have not worked at all (they import error during discovery causing a failure) so we don't have any result updates on the periodic jobs since then. This has also caused terra to break the APIs used in other benchmarks so even after fixing the import error so things can run again things fail.

mtreinish:
Ok, this should be good to go now (assuming I got all the pylint errors). Everything is running on current master (with the exception of tomography which skips because it'll have to be rewritten using ignis. But that'll have it's own class of backwards compat problems, because ignis only works with one version of terra at a time) for me locally now and I think I got the backwards compat shims all setup properly so it will use the old API for older versions of terra.

jaygambetta (approved):

PR #289, opened 2019-05-23
Update getting_started.rst

jaygambetta:


derivation:
@jaygambetta we were going back and forth on whether to use the new register-less syntax. We'll default to this for simplicity going forward.

derivation (approved):

PR #295, opened 2019-05-23
update install instructions

### LINKED ISSUE #288: update the installing
1. we should make sure we are starting to direct to the new quantum epxerience 

image

2. we should make sure that now all the element are part update the version 

image

derivation:
closes #288

- [x] look into how to revise section with special instructions for IBM Q Network members


jaygambetta:
you need the url for both

jaygambetta (approved):

PR #302, opened 2019-05-24
Bumping up Aer patch version

atilag:


mtreinish (changes requested):
Thanks, LGTM. But can we also add this to the changelgog too?

mtreinish (approved):

PR #304, opened 2019-05-24
Remove workaround for upgrading from qiskit < 0.7

### LINKED ISSUE #271: Remove the workaround for upgrading from terra <0.7
### Informations

- **Qiskit version**: >=0.7
- **Python version**: Any
- **Operating system**: Any

When installing the qiskit meta-package we have a workaround that relies on pip internals to force alternate commands when upgrading from qiskit<0.7 to qiskit-terra>=0.7. This is because in the 0.7 release we switched to a namespace package and split terra and aer. The qiskit metapackage owns the qiskit package name, but terra still owns the python namespace. However these workarounds involve forcing the wheel build to fail which leads to confusing errors being printed:
```
Building wheels for collected packages: qiskit
  Building wheel for qiskit (setup.py) ... error
  ERROR: Complete output from command /tmp/fun/bin/python3 -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-req-build-hn0ryofy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-ba18zhoz --python-tag cp37:
  ERROR: running bdist_wheel
  ----------------------------------------
  ERROR: Failed building wheel for qiskit
  Running setup.py clean for qiskit
Failed to build qiskit
Installing collected packages: qiskit
  Running setup.py install for qiskit ... done
Successfully installed qiskit-0.10.0
```

The lines starting with error are also printed in red on color terminals. This is very confusing for users and we get at least 1 person per week trying to install qiskit who is concerned about this.

`pip install qiskit`

No error messages during that command

### Suggested solutions

It's more a discussion about when we can remove these workarounds. At some point we can just say we don't support pip upgrades from <0.7.0 to anything newer. If you're using an old version uninstall and reinstall manually with:

`pip uninstall qiskit`
`pip install -U qiskit`

mtreinish:
### Summary

When we switched to the qiskit package being a metapackage instead of
terra in the 0.7 release there were issues introduced by upgrading from
the pre-0.7 terra package to the metapackage that would result in a
broken install. We were able to come up with a workaround that involved
shelling out to pip to manually reinstall terra during the install
process and forcing the wheel build to fail to ensure we always
installed from sdist. However these workaround caused a certain degree
of headache for users because pip will display an error despite it not
being fatal and expected. Now that we're several releases past 0.6.x
the importance of providing a direct upgrade path is not as important
since most users will be coming from at least 0.7. The value this
provides is not as critical anymore and it's frankly outweighed now by
the user confusion. This commit removes the workarounds and documents
the manual uninstall needed for upgrading from a qiskit version < 0.7.0.

### Details and comments

Fixes #271


mtreinish:
I'm marking this as on hold until we're ready to push out a `0.11.0` release, since including this in a `0.10.x` patch version would be a bit confusing.

mtreinish:
I've removed the on hold since 0.11 will be the next release: https://github.com/Qiskit/qiskit/pull/364

atilag (approved):

PR #315, opened 2019-06-08
Added benchmarks for isometries

rabaniten:
Added benchmarks for the run-time and the gate counts for the decomposition of isometries from 
m to n qubits for different values of m and n.


test/benchmarks/isometry.py
```diff
@@ -0,0 +1,35 @@
...
+    params = ([0, 1, 2, 3, 4, 5, 6], [1, 2, 3, 4, 5, 6])
+    param_names = ['number of input qubits', 'number of output qubits']
+
+    def setup(self, m, n):
+        iso = random_unitary(2 ** n).data[:, 0:2 ** m]
+        if len(iso.shape) == 1:
+            iso = iso.reshape((len(iso), 1))
+        q = QuantumRegister(n)
+        qc = QuantumCircuit(q)
+        qc.iso(iso, q[:m], q[m:])
```
mtreinish, line 36: This depends on Qiskit/qiskit-terra#2600 having merged right? Can we add:

```
if not hasattr(qc, 'iso'):
    raise NotImplementedError
```

before this line? That way when someone tries to run the benchmarks on an earlier commit this will just be skipped.

mtreinish (changes requested):
Thanks for writing this, it LGTM for the most part. Besides the lint failure, we should also detect whether the `iso` method is present or not on the quantum circuit object. That way this benchmark will only be attempted to run on commits after the dependent PR in terra lands (the benchmarks are run on current master as well as at any point in the git log for comparison purposes). I think this wi'll be ready to go after that.

test/benchmarks/isometry.py
rabaniten: Thanks a lot for the prompt feedback! I have updated it as you suggeted (and fixed some other minor things)

test/benchmarks/isometry.py
```diff
@@ -0,0 +1,44 @@
...
+            raise NotImplementedError
+        qc.iso(iso, q[:m], q[m:])
+        self.circuit = qc
+
+    def time_simulator_transpile(self, *unused):
+        transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'])
+
+    def track_gate_counts(self, *unused):
+        circuit = transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'])
+        return circuit.count_ops()
```
mtreinish: The docs say the return for a generic tracking benchmark are numeric (https://asv.readthedocs.io/en/stable/writing_benchmarks.html#tracking-generic), but I wasn't sure if a dict of numeric values would count. I just ran this locally to check and it does not work. While the benchmarks run fine, when it comes to aggregating the results for publishing, it fails when it tries to do numeric analysis (well if an average counts as analysis :p) on the results. So we either need to change the result to be a single number, or change the metric we're tracking to be a single number per execution.

mtreinish (changes requested):
I think these benchmarks are good, and I was ready to approve it. But, we need to change the return format for `track_gate_counts()`. This won't work while returning a dictionary.

test/benchmarks/isometry.py
rabaniten: Thanks for noting this. I have updated the following:
-We consider the number of C-NOT gates as a cost measure (the number of single-qubit gates is anyway bounded by about 4 times the number of C-NOTs, and the C-NOTs are usually much more expensive to implement)
-I have updated the parameter range to avoid evaluating non valid isometries (we must have n>=m for an isometry from m to n qubits)

test/benchmarks/isometry.py
```diff
@@ -0,0 +1,49 @@
...
+        self.circuit = qc
+
+    def time_simulator_transpile(self, *unused):
+        transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'])
+
+    def track_cnot_counts(self, *unused):
+        circuit = transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'])
+        counts = circuit.count_ops()
+        if 'cx' in counts.keys():
+            cnot_count = counts['cx']
```
mtreinish: We can replace L45-L48 with `cnot_count = counts.get('cx', 0)` which will use 0 as a default if cx is not defined in the dictionary.

mtreinish (changes requested):
Thanks for the quick update. There's just one small quick inline code simplification I think we should make here. After that small change is made this LGTM and I'll merge.

mtreinish (approved):
I took care of the simplification I mentioned before, and also set a seed for all functions that took one to reduce noise in the benchmarks. Also moved the skip raise as early as I could in the `setup()` (which are all things I noticed on a second look). With this it's ready to merge now.

PR #336, opened 2019-06-19
Typos in Docs

### LINKED ISSUE #335: Typos / Broken Links in Docs
This issue has been created to solve the typos that has been found by the Japanese translators during translation and to update the broken links.

SooluThomas:
fixed #335

### DIFF
--- a/docs/terra/visualizing_a_quantum_circuit.rst
+++ b/docs/terra/visualizing_a_quantum_circuit.rst
@@ -123,7 +123,7 @@ Controlling output from circuit.draw()
 
 By default the draw method returns the rendered image as an object and
 does not output anything. The exact class returned depends on the output
-specified: ``'text'``\ (the default returns a ``TextDrawer`` object,
+specified: ``'text'``. The default returns a ``TextDrawer`` object,
 ``'mpl'`` returns a ``matplotlib.Figure`` object, and ``latex`` returns
 a ``PIL.Image`` object. Having the return types enables modifying or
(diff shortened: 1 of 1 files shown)


lzdanski (approved):
LGTM

docs/terra/visualizing_a_quantum_circuit.rst
```diff
@@ -123,7 +123,7 @@ Controlling output from circuit.draw()
 
 By default the draw method returns the rendered image as an object and
 does not output anything. The exact class returned depends on the output
-specified: ``'text'``\ (the default returns a ``TextDrawer`` object,
+specified: ``'text'``. The default returns a ``TextDrawer`` object,
```
mtreinish, line 126: While this was a typo, this fix doesn't actually make it read well. The default it is talking about here is the text output type, not the return. So rewording it as a separate sentence like this makes it unclear about what is the default.

My intent here when I originally wrote the notebook this is based on was to have "the default" in parenthesis to mark the text output type. So it should read:

`The exact class returned depends on the output specified: ``'text'`` (the default) returns a ``TextDrawer`` object,`

Or if I was rewriting it today I'd probably make it:

`The exact class returned depends on the output specified: ``'text'`` (the default output type) returns a ``TextDrawer`` object,`

SooluThomas: Thank you @mtreinish! I'll update it now.

PR #343, opened 2019-06-21
Add first benchmarks for circuit construction and composition.

kdk:
Add two benchmarks timing construction of a simple circuit of one and two qubit gates at various widths and lengths, and another for composing the same circuit on top of an empty circuit.


ajavadia (dismissed):

mtreinish (approved):
Thanks for this, LGTM.

PR #344, opened 2019-06-21
Random circuit hex less random

kdk:
Attempt to remove some of the randomness in the random_circuit_hex benchmark ( https://qiskit.github.io/qiskit/#random_circuit_hex.BenchRandomCircuitHex.track_depth_ibmq_backend_transpile ) by seeding the random_unitary generation and the transpiler.


mtreinish (commented):
LGTM, hopefully this will improve some of the data we get from these benchmarks. One small thing, but I'm thinking a change like this warrants bumping the version of the benchmark (see L77). Since it could arguably produce different results from previous runs. What do you think?

kdk:
I think that makes sense. There's not a lot of signal in the old data, so I don't think much is lost in bumping the version number, and we can always requeue old commits.

jaygambetta (approved):

PR #348, opened 2019-06-24
Add the Qiskit/qiskit and ibmq provider to generate_authors.py

mtreinish:
### Summary

We were not tracking the contributions to the Qiskit/qiskit
meta-repository which is where we host all the documentation,
benchmarks, and anything else that is used between elements or
the Qiskit/qiskit-ibmq-provider . This commit corrects the oversight
adds the both to the list of repositories to use for generating the
authors list.

### Details and comments

With Qiskit/qiskit-aqua#592 having merged now,
the qiskit-chemistry git log has merged into the qiskit-aqua repository.
This means we no longer need to parse the log for both aqua and
chemistry since just aqua will contain all the git history for both.


jaygambetta:
why remove https://github.com/Qiskit/qiskit-chemistry

mtreinish:
I removed chemistry because https://github.com/Qiskit/qiskit-aqua/pull/592 merged the git history of chemistry into aqua. So we only need the aqua repository to track authorship now.

jaygambetta (approved):

PR #352, opened 2019-06-27
Update LICENSE.txt

anamita:
updated date


jaygambetta:
@mtreinish how do you want to do this for changing years.

mtreinish:
Hmm, if we do want to update it I'd probably say do it as 2018-2019. The work we did in 2018 in this repo is still under the same copyright.

jaygambetta:
yeah I dont think we update -- I think the rule is the date and if we change the file we can add 2018-2019

PR #358, opened 2019-07-02
random_circuit_hex benchmark backwards compatility

kdk:
Makes two additional changes to random_circuit_hex to better support older qiskit versions.

1. Re-orders import attempts for `euler_angles_1q`. At some point in time, trying to call `euler_angles_1q` via `qiskit.mapper` would import successfully, but raise when called with an error like `qiskit.exceptions.QiskitError: 'euler_angles_1q functionality is now accessible in qiskit.quantum_info.synthesis'`. Change default order to always use the newest version of `euler_angles_1q` that's available.

2. Falls back to setting `seet_mapper` for versions of qiskit earlier than `seed_transpiler`.


mtreinish (approved):
This LGTM, hopefully this will address all the compat issues moving between 0.7..HEAD

PR #306, opened 2019-05-26
randomize benchmarking performace benchmark

ShellyGarion:
### Summary
#272

PR #361, opened 2019-07-05
update qft benchmark: add Aer timing, go up to 24 qubits

ajavadia:
updates to QFT benchmark

### DIFF
--- a/test/benchmarks/qft.py
+++ b/test/benchmarks/qft.py
@@ -17,18 +17,20 @@
 
 import math
 
-from qiskit import QuantumRegister, QuantumCircuit
-from qiskit import BasicAer
+from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
+from qiskit import Aer
+
 try:
-    from qiskit.compiler import transpile
+    from qiskit.compiler import transpile, assemble
 except ImportError:
     from qiskit.transpiler import transpile
+    from qiskit.converters import circuits_to_qobj as assemble
 
 
-def build_model_circuit(qreg, circuit=None):
+def build_model_circuit(qreg, creg, circuit=None):
     """Create quantum fourier transform circuit on quantum register qreg."""
     if circuit is None:
-        circuit = QuantumCircuit(qreg, name="qft")
+        circuit = QuantumCircuit(qreg, creg, name="qft")
 
     n = len(qreg)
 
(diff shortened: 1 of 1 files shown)


test/benchmarks/qft.py
```diff
@@ -17,46 +17,59 @@
 
 import math
 
-from qiskit import QuantumRegister, QuantumCircuit
-from qiskit import BasicAer
+from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
+from qiskit import Aer
```
mtreinish, line 21: If we're going to execute the circuit on aer now too we'll have to install it in the benchmark env. Can you add a commit like: https://github.com/Qiskit/qiskit/pull/306/commits/b03f6744d47d22b2f601e2cc2eb019dd335b2aa2 to make sure we install aer in the asv environment. That being said these benchmarks track terra's performance over time, we'll always be testing the circuit on a fixed version of aer because of that, is there still value in benchmarking the execution?

test/benchmarks/qft.py
```diff
@@ -17,46 +17,59 @@
...
     for i in range(n):
         for j in range(i):
             circuit.cu1(math.pi/float(2**(i-j)), qreg[i], qreg[j])
         circuit.h(qreg[i])
+    circuit.measure(qreg, creg)
 
     return circuit
 
 
 class QftTranspileBench:
```
mtreinish, line 46: We should add a version parameter here to make it explicit that the results here are different from previous executions of the qft benchmarks. Just add a parameter to the class:

`version = 2`

mtreinish (changes requested):
LGTM, but besides the lint failure there are a couple inline comments. Mostly we need to version the benchmark now to mark it as different from previous results. Also, if we're going to execute circuits in aer we'll need to make sure it's installed. But, it raises bigger questions about what we're benchmarking there, if we're trying to measure aer performance over time this probably isn't the right place to do it.

jaygambetta:
@ajavadia and @mtreinish did you close on this.

mtreinish:
@jaygambetta there hasn't been any update to this since my comments last month. I assume it's lower priority and @ajavadia hasn't had a chance to look at it more

mtreinish:
I'm going to close this since there hasn't been any recent activity. Feel free to reopen it if you'd like to pick this up again.

PR #309, opened 2019-05-29
Bump terra and prepare for release

mtreinish:
### Summary

This commit bumps the terra version to pull in the latest release 0.8.1.
It also prepares the qiskit package versions for release.

### Details and comments

Fixes #278

PR #364, opened 2019-07-10
Update Aer and IBMQ Provider dependencies and bump up Qiskit metapackage version to: 0.11.0

atilag:
### Summary
* Update qiskit-aer version to 0.2.2
* Update qiskit-ibmq-provider to 0.3.0
* Bump up qiskit package to 0.11.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,18 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 > -   **Fixed**: for any bug fixes.
 > -   **Security**: in case of vulnerabilities.
 
-[UNRELEASED](https://github.com/Qiskit/qiskit-terra/compare/0.10.5...HEAD)
+[UNRELEASED](https://github.com/Qiskit/qiskit-terra/compare/0.11.0...HEAD)
 ==========================================================================
+
+[0.11.0](https://github.com/Qiskit/qiskit/compare/0.10.5...0.11.0) - 2019-07-11
+===============================================================================
+
+Changed
+-------
+
+- Updated the `qiskit-aer` dependency to 0.2.2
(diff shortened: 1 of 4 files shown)


mtreinish (changes requested):
Can you also update the version list in docs/release_notes.rst and bump the docs version in docs/conf.py. You can see an example of this in: https://github.com/Qiskit/qiskit/pull/309

We also probably should add a 0.11 entry in the release notes proper just to outline why it's a minor version (ie the ibmq provider changes).

chriseclectic:
@atilag, @quantumjim found a significant bug in Aer 0.2.2. It should be fixed in [this PR](https://github.com/Qiskit/qiskit-aer/pull/280).

Can we merge that in and bump Aer to 0.2.3 before updating the metapackage?

nonhermitian:
The `qiskit-ibmq-provider ` has a bug where updating to the new API causes a loss of job history: https://github.com/Qiskit/qiskit-ibmq-provider/issues/248.

I would wait until that is fixed to bump the version (unless the new API is not included).

diego-plan9:
> The qiskit-ibmq-provider has a bug where updating to the new API causes a loss of job history

For context, it actually seems to be the effect of a change introduced inadvertently in the API in the latest deploy, which the team is working upon and seems an easy fix on their end. Updating to the new API will not cause a loss of job history, nor changes seem to be needed in the provider itself, so hopefully we can proceed while keeping an eye on the API.

nonhermitian:
I would like to verify the jobs are returned before this goes out. Also a few items to do before this can be released.

nonhermitian:
I believe that this is just waiting on the provider to internally grab all the jobs requested by the limit kwarg and then it is ready.

diego-plan9 (approved):

PR #366, opened 2019-07-11
Adds Aer benchmarks

atilag:
### Summary
Adds `Aer` benchmarks
These are exactly the same benchmarks we are running in a daily basis in a dedicated baremetal machine.


kdk:
Maybe set `optimization_level=0` and `seed_transpiler` in calls to `transpile` so that changes to transpiler optimizations or RNG don't show up as noise in your benchmarks?

atilag:
oh, ok. Makes sense.

atilag:
As a side-channel conversation, I'm going to remove transpilation step, we don't really need it for running the benchmarks, and yeah, it can interfere in many unwanted ways.

atilag:
@chriseclectic told me that we do need to unroll arbitrary 2-qubit gates to single qubit gates and CX gates, and this decomposition happens in the transpilation, so I'm reintroducing it but removing optimizations as @kdk said, and setting a fixed value to the seed.

PR #396, opened 2019-08-08
Add single-threaded versions of RB benchmarks.

kdk:
`qiskit.transpile` will attempt to distribute circuits for transpilation across any available cores, generally reducing transpile times, but measuring single threaded performance can provide more reliable insight into how individual commits impact performance and will allow for measuring the benefit of parallel vs serial transpilation.

This PR adds a single-threaded version to the RB benchmark, which currently transpiles in parallel, by hard coding the `QISKIT_IN_PARALLEL` environment variable.


mtreinish (changes requested):
The idea behind this change LGTM, but I'm concerned about the environmental variable use here. Will the variable stick between runs of the benchmark? I think just to be safe we should add a `teardown`[1] method to both benchmark classes to make sure we unset the env variable (if it's set) after each benchmark execution.

[1] https://asv.readthedocs.io/en/stable/writing_benchmarks.html#setup-and-teardown-functions

jaygambetta:
@mtreinish did you and @kdk close on this.

mtreinish:
@jaygambetta @kdk and I were talking about this one the other day. There are some last minute changes he was working on before this is ready.

mtreinish (approved):

PR #397, opened 2019-08-08
Documentation source code linebreaks

### LINKED ISSUE #394: Documentation source code linebreaks
The current documentation is not wide enough to contain the full lines of source code. This causes many line breaks and makes reading the source code on the documentation website unworkable. Since the max line width in the linters is 100 characters, the source documentation should show at least 100 characters per line to prevent linebreaks.

See, for example, https://qiskit.org/documentation/_modules/qiskit/aqua/operator.html#Operator.construct_evaluation_circuit

SooluThomas:
fixed #394

### DIFF
--- a/docs/theme/static/css/theme.css
+++ b/docs/theme/static/css/theme.css
@@ -203,7 +203,7 @@ a.reference.external:hover{
 .document{
     min-height: calc(100vh - 230px);
     background: white;
-    max-width: 1000px !important;
+    max-width: 1200px !important;
     margin: 0 !important;
 }
 


jaygambetta (approved):

eddieschoute:
This does not resolve the issue. The source needs to show _at least_ 100 characters not at most `x` pixels. In my opinion, it would be preferred to have a horizontal scrollbar over linebreaks which is why I say to show at least 100 characters.
Fullscreen on a new macbook pro the lines still wrap. Related to this is that the font size is currently 16px. I think this could be reduced to maybe 14px. I think we should consider aiming for no line wrap nor scrollbar at a standard 1080p fullscreen window if that's possible.

SooluThomas:
@delapuente Can you please look into this issue?

PR #399, opened 2019-08-12
redirecting Advanced Use of IBM Q Devices

### LINKED ISSUE #290: do we need https://qiskit.org/documentation/advanced_use_of_ibm_q_devices.html
my feeling is we dont need. this page. If we do keep it we should make it much simpler it does not need to mimic the notebook

derivation:
### Summary
closes #290 

### Details and comments

- [x] redirect links to this page to the tutorial
- [x] remove this page from the doc table of contents
- [x] remove the source file for this page


mtreinish (commented):
I don't think this is a full solution for a redirect. If a user has the docs page https://qiskit.org/documentation/advanced_use_of_ibm_q_devices.html bookmarked, in their history, on a stale search result page, etc when this merges that like will no longer work. The user will have no idea where the page went just that they got a 404.

I feel like we should either retain this page and simply just provide a link to the old one. Or setup a redirect on the webserver that the documentation is hosted from so that it links to the tutorial when someone goes to the current page.

derivation:
Excellent point, @mtreinish

derivation:
- [x] redirect links in doc to the tutorial
- [x] remove this page from the doc table of contents
- [x] remove the source file for this page
- [ ] setup a 301 redirect on the webserver

Add the following to .htaccess
`Redirect 301 /old-page/ http://www.yourdomain.com/new-page/`

jaygambetta (approved):

PR #404, opened 2019-08-14
Add details on qiskit-terra warnings to install guides

mtreinish:
### Summary

This commit adds details to the install guide and the install from
source guide on new warnings being emitted by terra in the 0.9 release.
Added in Qiskit/qiskit-terra#2842 when qiskit-aer or
qiskit-ibmq-provider can not be found by terra on initial import it will
emit a warning to inform the user of that state. This is done because
today the user will have no idea if they forgot to install these
components (or more likely python packaging screwed up the installation)
and aer or ibmq can't be found. They will then try to use either
qiskit.Aer or qiskit.IBMQ only to be met by a confusing ImportError.
This is a very common occurrence for users and causes a great deal of
confusion. However, since emitting a warning on import is potentially
undesirable behavior for any users that only use qiskit-terra by
itself. If they never have any intent of importing qiskit.Aer or
qiskit.IBMQ then emitting a warning is fairly annoying. This commit
documents these warnings and more importantly how to suppress them for
these users, so that they do not have to be concerned about seeing the
warnings in perpetuity moving forward if they do not want to.

This updates both the install guide and the contributing guide since
it's conceivable a user would refer to either guide when installing
terra as a standalone component.

At the same time a stale reference to a warning that as of #304 merging
no longer will be emitted when installing the qiskit metapackage is
removed from the install documentation.

### Details and comments

### DIFF
--- a/docs/contributing_to_qiskit.rst
+++ b/docs/contributing_to_qiskit.rst
@@ -252,6 +252,23 @@ Once the compilers are installed, you are ready to install Qiskit Terra.
 
 After you've installed Terra, you can install Aer as an add-on to run additional simulators.
 
+.. note::
+
+    If you do not intend to install any other components qiskit-terra will
+    emit a ``RuntimeWarning`` warning that both qiskit-aer and
+    qiskit-ibmq-provider are not installed. This is done because the more
+    common case is to have users who intend to use the additional elements
+    but not realize their not installed, or have the installation fail. If
+    you wish to suppress these warnings this is easy to do by adding::
+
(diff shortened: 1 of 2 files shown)


mtreinish:
This is also set to be included in the terra release notes here: https://github.com/Qiskit/qiskit-terra/pull/2919/commits/b768804c968736b4965a38277023be59bfa9793b as part of Qiskit/qiskit-terra#2919

jaygambetta (approved):

Issue #390, opened 2019-07-31
Standardize the links in .rst files


SooluThomas:
## Why standardize?

This standardizing is required to adapt a common way to write the docs and also to avoid confusion for the translators regarding different syntaxes.

## Changes needed

- There is need for standardizing the cross-referencing links from 
   ```
   `link`_

   .. _link: <http://www.example.com>
   ```
   to 
   ```
   `link <http://www.example.com>`__
   ```
   [This](https://github.com/Qiskit/qiskit/pull/383/commits/e388742300ee178d2c1e54bd81f7622f7a62b82d) is an example commit of editing the cross-referencing links

This change is required because the translators can't translate "link" to the target language because "link" itself is a pointer to a link. Inline links can solve this problem.

- Also there is a difference between single underscore and a double underscore in links:

   Single underscores are for **non-anonymous** links. For example:
   ```
   `link <www.example1.com>`_
   `link <www.example2.com>`_
   ```
   This will generate a warning from sphinx:
   `(WARNING/#) Duplicate explicit target name:link`
   because `link` points to two different locations.
   For more details on where single underscores are used, check this [link](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#hyperlink-references)

   Having double underscores are for **anonymous** links,
   ```
   `link <http://www.example1.com>`__
   `link <http://www.example2.com>`__
   ```
   will generate 
   ```
   <ol>
   <li><a href="http://www.example.com">link</a></li>
   <li><a href="http://www.example.com">link</a></li>
   </ol>
   ```
   which will avoid such warnings.
   For more info on this, check this [link](https://stackoverflow.com/questions/5464627/how-to-have-same-text-in-two-links-with-restructured-text)

So, its recommend to use **double underscore** instead of single ones to avoid such warnings

@lzdanski @derivation your thoughts?

PR #411, opened 2019-08-17
Docs cleanup

nonhermitian:
### Summary
- Removed all content that overlapped with tutorials (Did not do Aqua yet)
- All elements highlighted on same page, and redundant text on main page removed.
- Changed name to highlight that this is API documentation
- Note saying this is API documentation and here is where tutorials are.
- Added link to tutorials in sidebar.


nonhermitian:
I had to temporarily disable line length checking because the links to the tutorials are very long, and I am unclear as to how to break rst links across multiple lines.

mtreinish:
> I had to temporarily disable line length checking because the links to the tutorials are very long, and I am unclear as to how to break rst links across multiple lines.

Try not using an inline link. Do it like the example in the code box here:

http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

in this case something like:
```
see the `Qiskit tutorials`_. A link can also be found in the sidebar.

.. _Qiskit tutorials: https://nbviewer.jupyter.org/github/Qiskit/qiskit-tutorials/blob/master/qiskit/1_start_here.ipynb

```

jaygambetta:
im approving even though the links need work and same with the elements. As well as the combining of the developer docs

jaygambetta (approved):

nonhermitian:
To be clear the link is too long no matter where it goes. I think maybe rst substitution several times can also fix it. 

This is a reasonable stop point before thinking of moving all element info into the source code. Perhaps that should be first thing on the list when we do migration.

derivation:
@nonhermitian try [this sphinx extension](https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html) for the crazy long tutorial links. It'll involve editing [this section](https://github.com/Qiskit/qiskit/blob/c4a1bde76576e1fb4a8b99e384c63a3af23506f8/docs/conf.py#L113) of the Sphinx conf.py.

SooluThomas:
> > I had to temporarily disable line length checking because the links to the tutorials are very long, and I am unclear as to how to break rst links across multiple lines.
> 
> Try not using an inline link. Do it like the example in the code box here:
> 
> http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links
> 
> in this case something like:
> 
> ```
> see the `Qiskit tutorials`_. A link can also be found in the sidebar.
> 
> .. _Qiskit tutorials: https://nbviewer.jupyter.org/github/Qiskit/qiskit-tutorials/blob/master/qiskit/1_start_here.ipynb
> ```

Inline links can help in better translation. It's better to have inline links so that the names of references can be translated and will point to the same required pages/link targets. More details are available in [GitHub Issue](https://github.com/Qiskit/qiskit/issues/390) #390 .

PR #421, opened 2019-08-17
fixing broken link

jaygambetta:

PR #422, opened 2019-08-17
Removed tutorials from TOC

SooluThomas:


jaygambetta:
This lgtm  But I would like @nonhermitian and @derivation to decide if we remove it here or in the header. If in the header we also need to remove from the qiskit.org repository.

jaygambetta:
approve as @nonhermitian commented on previous pr #421 that he prefers this way

jaygambetta (approved):

PR #427, opened 2019-08-18
Terra API Cleanup

nonhermitian:
### Summary
Now is collected as follows:

Screen Shot 2019-08-18 at 6 36 21 AM

Screen Shot 2019-08-18 at 6 36 34 AM

### Details and comments
I can get all of terra done, but may need help with other elements.


jaygambetta:
I would not split them by element in the api documentation

nonhermitian:
I did it so that qiskit.aqua main folders would be at same level as terra ones. But can do as you suggest. This also means all of aer is hidden in providers.

jaygambetta:
I think it is ok for aer to be hidden in providers. Then all the providers, ibmq, basicaer, and aer are the same level. In the providers we need some source code that gives the concept of providers. 

Ignis will have one module which has sections which I think @dcmckayibm is good. David do you like qiskit.Ignis.Characterization qiskit.Ignis.verification, qiskit.Ignis.mitigation or would you prefer qiskit.characterization, qiskit.verrification, qiskit.mitigation. My vote would be the first and then in qiskit.ignis you can have a bit more about the fitters and filters in the source code that gives the main reasoning, but David if you want to promote the three main modules to a higher level I am good. 

Aqua yeah I think we need to talk to the team a bit more. Looking at the current they have chemistry  at the level of aqua at the moment. There would be qiskit.aqua, qiskit.chemistry as the main modules  and if they bring in qiskit.optimzaiton and qiskit.finance and qiskit.ai (I forget if you want ai or artificial intelligence, I don't mind) aqua will bring in 5 major modules into qiskit which I think is good (the terra team just needs to make sure that done make these names in the main repo).  @pistoia I will let you decide how you want the auto folders, but my vote is to have the 5.

nonhermitian:
Explaination in the module docs for providers

nonhermitian:
@jaygambetta I have basically all of the Terra API covered here.  Should look much nicer.  Will make issues for other elements and provider for docs team.

jaygambetta:
ok I will approve just so we have a reference and lets go from here. Looks good to what I can see.

jaygambetta (approved):

pistoia:
@jaygambetta I am in favor of the 5 as well.  It gives more visibility to the various domains instead of favoring just chemistry.  I am planning to make this change for the next release (Aqua 0.6.1) because we are waiting for a new substantial contribution to Qiskit Optimization, which will make the Optimization stack resemble the Chemistry stack.

jaygambetta:
I agree

dcmckayibm:
I would do qiskit.ignis and that can be one document, no need to further break out

PR #447, opened 2019-08-19
Another shot at Terra docs

nonhermitian:
### Summary
Another shot at moving away from autodoc, and making things look good.

### Details and comments
@mtreinish can you please check the deploy script to verify that when this gets merged it does what it should.


jaygambetta:
you need to also fix the links in the_elements.rst and I think there are some in the Readme.md and contributing.md as well.

docs/contributing_to_qiskit.rst
```diff
@@ -295,21 +295,21 @@ universally depending on operating system.
...
 
          .. code:: sh
 
-            sudo apt install build-essential
+            dnf install @development-tools
 
       4. Install OpenBLAS development headers.
 
-         If you're using Ubuntu>=16.04 or an equivalent Debian Linux distribution,
+         If you're using Fedora or an equivalent Debian Linux distribution,
```
mtreinish: Small nit: we probably should remove the debian reference here

```suggestion
         If you're using Fedora or an equivalent Linux distribution,
```

nonhermitian: ahh missed one

docs/conf.py
```diff
@@ -57,6 +57,7 @@
     'sphinx.ext.viewcode',
     'sphinx.ext.extlinks',
     'sphinx_tabs.tabs',
+    'sphinx_automodapi.automodapi'
```
mtreinish:
```suggestion
    'sphinx_automodapi.automodapi',
```

mtreinish:
With the inline suggestion above I think this will work as expected. I locally tested the deploy script from a fresh venv (which is similar to a clean travis env) and ran it like travis would and it generated this commit:

https://github.com/mtreinish/qiskit.org/commit/0a3076d8a98f44146284990e32fa6c7773b069f7

which looks like it includes the updated api docs for terra

nonhermitian:
Change integrated.

mtreinish (approved):
I think this should be good to go now. If there are any issues with the deploy script I wasn't able to find them locally and we can debug them post merge.

PR #455, opened 2019-08-19
Cleanup old referencing labels from docs

### LINKED ISSUE #453: Cleanup old referencing labels from docs
Remove all no-longer-used referencing labels. 

For example:
Screen Shot 2019-08-19 at 10 28 42 AM

SooluThomas:
### Summary

fixes #453

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -186,8 +186,8 @@ Step 4 : Visualize the Circuit
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can use ``QuantumCircuit.draw()`` to view the circuit that you have designed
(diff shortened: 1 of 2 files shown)


SooluThomas:
Quick question: What is actually referenced by the label:
```
:ref:`top-level functions <qiskit_top_level_functions>`
```

Screen Shot 2019-08-19 at 12 10 13 PM

nonhermitian (approved):

jaygambetta:
I dont know if we want to drive to the tutorials. Can we minimize the link and if we need them link to the API source code like I did in the qiskit elements file.

SooluThomas:
> I dont know if we want to drive to the tutorials. Can we minimize the link and if we need them link to the API source code like I did in the qiskit elements file.

okay. I'll do the necessary changes.

PR #463, opened 2019-08-19
Add timing benchmark for `import qiskit`.

kdk:
Adds a first benchmark to measure how long it takes to import top-level qiskit. This will include some overhead in starting up a new python interpreter, but that should be relatively consistent. ( https://github.com/numpy/numpy/pull/8159 has some discussion over their approach, which is the same as is taken here.)


mtreinish (dismissed):
LGTM, a nice simple benchmark and important to track. Especially because I've added several things that I know make the import slower (like the `__qiskit_version__` pip checks). It does make me wish there was a good way to track the import time in an isolated venv (like what asv gives us) vs a full environment with all the elements installed so the namespace is larger.

jaygambetta (approved):

PR #498, opened 2019-08-26
Remove toggle button

SooluThomas:
This PR will remove the language toggle button till the translation workflow is fixed.


mtreinish (approved):
Normally I prefer to just delete things we're temporarily removing, a git revert is just as easy as removing a comment. But either way is good enough I guess

Issue #475, opened 2019-08-21
Enable Doc Translation


SooluThomas:
- [x] Create a branch for pushing in .po (portable object) files.
- [x] Update .travis.yml and deploy_documentation.sh to configure the translation build process.
- [x] Update Qiskit styling for locale.
- [x] Add .po files/folder to new branch.
- [x] update .gitignore of master and poBranch
- [x] add a LOCALIZATION_CONTRIBUTORS file 
- [x] add info on how to contribute to localization in docs
- [x] connect Qiskit-Docs Crowdin project to poBranch

derivation:
@lzdanski if you get me a bullet point version of the envisioned CLA workflow, I can make sound good.

derivation:
So, barring something unexpected coming up with the last legal review, I'd say this is mission complete. Great job everyone! Next steps are to promote the translation workflow to community members. Yuri is working on a blog post to kick that off.

PR #502, opened 2019-08-27
Automated po file generation

SooluThomas:
Updating some code changes from issue https://github.com/Qiskit/qiskit/pull/489 because it is breaking travisCI build.

#475


derivation (approved):
Spoke with Soolu about this. Looks good.

PR #510, opened 2019-08-27
Swap order of keys for ssh-add in deploy_documentation

mtreinish:
### Summary

This commit swaps the order we add keys to the ssh agent in
deploy_documentation script. It is currently failing on trying to push
to poBranch with a permission denied error, which is probably because
it's using the wrong key. This changes the script to verify that it
works when the only configured key is for the respective push.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
       sudo: true
     - python: "3.5"
       env: TOXENV=lint
-    - if: type = push
+    - if: branch = master AND type = push
       os: linux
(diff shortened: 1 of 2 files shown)


mtreinish:
This works now: https://travis-ci.com/Qiskit/qiskit/jobs/228653346 (and it pushed to poBranch and qiskit.org successfully) I'll remove the test patch and this will be ready to merge.

SooluThomas (approved):
LGTM! Thank you @mtreinish

PR #511, opened 2019-08-27
Add git merge to sync both master and poBranch

SooluThomas:
### Summary

This commit will sync both `master` and `poBranch` so that the translated files are in sync with the source .rst files (english). 

Referencing issue #475

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -26,22 +26,8 @@ SOURCE_REPOSITORY="git@github.com:Qiskit/qiskit.git"
 TARGET_BRANCH_PO="poBranch"
 DOC_DIR_PO="docs/locale"
 
-# Build the documentation.
-make doc
-
-echo "show current dir: "
-pwd
-
-cd docs
-
-# Extract document's translatable messages into pot files
-# https://sphinx-intl.readthedocs.io/en/master/quickstart.html
-echo "Extract document's translatable messages into pot files: "
-sphinx-build -b gettext -D language=$TRANSLATION_LANG . _build/gettext
-
-# Setup / Update po files
-echo "Setup / Update po files"
-sphinx-intl update -p _build/gettext -l en
+echo "git branch at the beginning"
+git branch
 
 echo "Setup ssh keys"
 pwd
@@ -61,6 +47,31 @@ cd temp
 git branch
 git config user.name "Qiskit Autodeploy"
 git config user.email "qiskit@qiskit.org"
+
+git checkout master
+git fetch origin/master
+git checkout $TARGET_BRANCH_PO
+git merge master
+git push origin $TARGET_BRANCH_PO
+
+echo "***************** Merged master with poBranch! **********************"
+
+# Build the documentation.
+make doc
+
+echo "show current dir: "
+pwd
+
+cd docs
+
+# Extract document's translatable messages into pot files
+# https://sphinx-intl.readthedocs.io/en/master/quickstart.html
+echo "Extract document's translatable messages into pot files: "
+sphinx-build -b gettext -D language=$TRANSLATION_LANG . _build/gettext
+
+# Setup / Update po files
+echo "Setup / Update po files"
+sphinx-intl update -p _build/gettext -l en
 
 echo "git rm -rf for the translation po files"
 # git rm -rf --ignore-unmatch $DOC_DIR_2/$TRANSLATION_LANG/**/*.po # Remove old po files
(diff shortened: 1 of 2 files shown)


SooluThomas:
test didn't work because the branch was made in my repo

tools/deploy_documentation.sh
```diff
@@ -62,6 +48,31 @@ git branch
 git config user.name "Qiskit Autodeploy"
 git config user.email "qiskit@qiskit.org"
 
+git checkout master
+git fetch origin/master
```
mtreinish: This isn't a valid command, `origin/master` is not a remote so you can't fetch it.

tools/deploy_documentation.sh
```diff
@@ -26,22 +26,8 @@ SOURCE_REPOSITORY="git@github.com:Qiskit/qiskit.git"
...
-# Extract document's translatable messages into pot files
-# https://sphinx-intl.readthedocs.io/en/master/quickstart.html
-echo "Extract document's translatable messages into pot files: "
-sphinx-build -b gettext -D language=$TRANSLATION_LANG . _build/gettext
-
-# Setup / Update po files
-echo "Setup / Update po files"
-sphinx-intl update -p _build/gettext -l en
+echo "git branch at the beginning"
+git branch
```
mtreinish: Why are you listing all the branches at the beginning?

mtreinish (changes requested):
I'm confused as to the purpose of this change, why do you need to merge master into poBranch? Isn't poBranch just where the generated .po files live. Shouldn't we just generate the English strings from master copy those into poBranch and push an update from that. What does merging master accomplish on top of that?

Additionally, trying to do a merge will have conflicts which must be resolved manually. Git is good at merging things automatically (using things like 3 way merge, etc), but there are a lot of cases where it can't resolve a conflict and manual intervention is required. This is especially true when the branches have diverged like master and poBranch have. This will fail as soon as you run it because there are merge conflicts (go ahead and try `git checkout origin/poBranch && git merge origin/master` in your local repo to see).

That all being said I found an issue inline too.

tools/deploy_documentation.sh
SooluThomas: to check what branch it is in at the beginning of the build. I found the answer by running a test build in my fork. I think I can remove `git branch`.

SooluThomas:
> I'm confused as to the purpose of this change, why do you need to merge master into poBranch? Isn't poBranch just where the generated .po files live. Shouldn't we just generate the English strings from master copy those into poBranch and push an update from that. What does merging master accomplish on top of that?

Both `master` and `poBranch` have to be in sync because, `ja` translations are built from both `.po` files and `.rst` files. Which requires updated `.rst` file in both `master` and `poBranch`.
(Either manual or automated).

For example, if there are any changes made to the [theme/layouts](https://github.com/Qiskit/qiskit/tree/master/docs/theme) in `master` branch, they won't be reflected in the finally generated translated html because the "Translated HTML" is generated from both the `.po` files and the `.rst` files in the repo/branch (wherever `sphinx-build -b html -D language=$TRANSLATION_LANG . _build/html/locale/$TRANSLATION_LANG` is run)

Here's a diagram from [Sphinx Internationalization Documentation](https://www.sphinx-doc.org/en/1.8/intl.html)
translation

Do let me know your views on this. Thank you.

tools/deploy_documentation.sh
SooluThomas: Thanks for pointing out the error. I'll rectify it.

mtreinish:
That diagram doesn't explain why you need to merge master into poBranch. In fact it doesn't mention branches anywhere, all it says is that the po and rst files are an input to sphinx build with a translated build. There is nothing that says this has to be in the same branch or even the same tree. Just that you need both.

The way things are setup now you should treat master as the source for the raw source (rst, layout etc) and poBranch as the source of the translated strings. Then when you build documentation (wherever you end up running that) you just pull from both sources. As I said earlier trying to merge master into branch automatically will not work and just prone to failure when there is divergence in the tree (which there already is). Things are actually already setup well to do this, on merges to master we generate new .pot files, update the .po files, and pushes that to poBranch already. That's all you need.

Then docs builds are as simple as something like::

```bash
git clone github.com/Qiskit/qiskit docs_source
git clone github.com/Qiskit/qiskit -b translations translations
cp -r translations/docs/locale docs_source/docs/.
sphinx-build ... docs_source/docs docs_source/_build
```
(although you don't actually need one of the clones because travis clones the repo for us)

SooluThomas:
That's a better idea. Thank you. I'll check if it can work that way.

SooluThomas:
Closing this issue because it's no longer valid

PR #513, opened 2019-08-28
Clone source files in master branch to generate updated html 

SooluThomas:
### Summary

This commit recorrects the logic of merging `master` branch to `poBranch` (mentioned in PR #511) to cloning source files in `master` (mentioned in the [comment](https://github.com/Qiskit/qiskit/pull/511#issuecomment-525832384)) to get updated source files and po files and generate translated html files that are consistent with English source html rather than having 2 sources.

referencing issue #475

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -34,11 +34,9 @@ ssh-add github_deploy_key
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone $SOURCE_REPOSITORY docs_source
-mkdir -p docs_source/docs/api && cp -r $SOURCE_DIR/docs/api/. docs_source/docs/api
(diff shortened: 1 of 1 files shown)


tools/deploy_documentation.sh
```diff
@@ -24,21 +24,24 @@ TRANSLATION_LANG='ja'
 # Build api .rst files
 make autodoc
```
mtreinish: Is this still needed if you're using master for the docs source? I thought we removed the autodoc pieces from the documentation build process as part of the recent refactor

mtreinish (approved):
LGTM, this is a much better approach to build the translated docs. One quick question inline. But it's outside the scope of the patch so we can always fix it after.

tools/deploy_documentation.sh
SooluThomas: I'll remove it

PR #519, opened 2019-08-29
Ignore LOCALIZATION_CONTRIBUTORS from authors file

mtreinish:
### Summary

The AUTHORS file is just for tracking those who have to the qiskit
project. However, changes to the docs/LOCALIZATION_CONTRIBUTORS file
in the Qiskit/qiskit repo are not actually contributions, it's a
workaround to enable translators to sign the cla. Translations are done
through another service, crowdin, however because translators need to
sign the CLA prior to contributing and crowdin has no such mechanism they
need to create a github PR to do so. This however is not actually an
indication of actual contribution just that they signed the CLA.
Translation contributions will be tracked separately using a different
mechanism eventually, but until then we don't want to accidently include
this CLA workflow in the AUTHORS file. This commit modifies the
generate_authors file to exclude this file from the git log used to
build the AUTHORS list.

### Details and comments

### DIFF
--- a/tools/generate_authors.py
+++ b/tools/generate_authors.py
@@ -33,9 +33,11 @@ def main(repos=None, output_path=None):
     authors = []
     emails = []
 
-    def generate_authors(git_dir):
+    def generate_authors(git_dir, meta=False):
         """Create AUTHORS file using git commits."""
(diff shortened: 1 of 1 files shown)


SooluThomas (approved):
Thank you @mtreinish

derivation:
@mtreinish thank you, this is exactly what we needed

PR #551, opened 2019-09-10
Build docs from individual source elements

nonhermitian:
### Summary
After moving the API docs to be back inside each element, this PR adds support for the meta-repository docs to leverage that. It adds a custom sphinx extension to pull in the source from each git repo at the version specified in the requirements list in the `setup.py` these apidocs then get copied into the sphinx source and the docs are built like before. This however will only happen after the local  `apidocs` dir is deleted from the the meta-repository (which requires each element to have a release including the api documentation). In the meantime you can leverage the env variable `DOCS_FROM_MASTER` by setting it to 1 (ie `DOCS_FROM_MASTER=1 tox -edocs`) and this will use each element from master (instead of at release) and build the docs the api docs. Which will also be useful in the future for testing.

### Details and comments
Files in Terra: https://github.com/Qiskit/qiskit-terra/pull/3085
Files in Ignis: https://github.com/Qiskit/qiskit-ignis/pull/269
Files in Aqua: https://github.com/Qiskit/qiskit-aqua/pull/678
Files in Aer: https://github.com/Qiskit/qiskit-aer/pull/339
Files in Account: https://github.com/Qiskit/qiskit-ibmq-provider/pull/338

Co-authored-by: Matthew Treinish <mtreinish@kortar.org>


nonhermitian:
The files in Terra: https://github.com/Qiskit/qiskit-terra/pull/3085

nonhermitian:
@SooluThomas this is a WIP for moving the sphinx source files into the elements.  @mtreinish is going to modify this PR to make the meta package docs work with source installed elements as well.  When the individual elements merge the docs in the above PR's we can start on actually doing the source code changes.

There is an additional step needed to glue all the pieces together that is not in this PR.

SooluThomas:
@nonhermitian @mtreinish We need to take necessary steps to make sure that localization build works and `.po` files of `apidocs` are generated.

nonhermitian:
Nothing should actually change in that respect.  The files would be copied into the `apidoc` directory and then built.

mtreinish:
I just pushed a first pass at the revision for this in: https://github.com/Qiskit/qiskit/pull/551/commits/d713ded4bff102085c67495873b7c1b148c8a4a4 I haven't fully tested all the different conditions for that yet but it's a good starting point. But, as I wrote in the commit msg for that I think we really should evaluate which models we support for building the docs from the meta repo. I only added the git piece because of @nonhermitian's comment in: https://github.com/Qiskit/qiskit-terra/pull/3085#issuecomment-529995914 but I'm not sure that's a model we really want to support.

My feeling is that each element's repo should be able to locally build the apidocs for that element (which is something we should test in CI). When doing development that's all that should be needed to test changes to apidocs. Each element's docs are self contained, they might reference an api in another element, but that's a static link and isn't needed for testing. Then the meta repo should aggregate the elements' apidocs for the releases given in the requirements file. The combined docs are really about released api not dev versions (which do change frequently and in backwards incompatible way inside of a release). If we want to test pre-release combined versions before release we can easily have a config flag to build from master on each element instead of release. But anything more than that I don't think is needed on the meta-repo side. But let me know your thoughts about that.

mtreinish:
Another option I just thought of is make the copy/clone path configurable with the model I described before. So that would enable testing a local branch in the integrated case but still keep the build model a bit cleaner

nonhermitian:
Yeah, so having thought about it more, if we have the ability to build the local elements docs (Which exists in the PR's I did) and any changes to meta can be built against whatever the latest pip is then we are set (provided the latest pip has the docs in them).

mtreinish:
Ok cool, I'll push up a revision shortly then to remove the from git version pieces (which will simplify it a lot).

mtreinish:
I think the only thing left to do here is to add a env variable flag to use master from each element instead of the releases (just for testing purposes). Then we can test this locally and start slowly phasing out the apidoc rst files in the meta package as the elements push releases with docs.

nonhermitian:
This works very well.

mtreinish:
@nonhermitian I think we need to update some of the conf and template stuff here because after the last commit I pushed on this branch (which was needed for this to work at all)  the build is throwing a ton of warnings like:

`/home/mtreinish/git/qiskit/qiskit/docs/apidoc/qiskit.rst:7: WARNING: toctree contains reference to document 'apidoc/transpiler' that doesn't have a title: no link will be generated`

for basically every module except for visualization. It also raises a larger question for me about the apidoc toctree in general. How do we ensure that every element's self contained docs end up in the toctree. Should we add a apidoc.rst file that has a 2 level toctree from `apidoc/qiskit` `apidoc/aer`, `apidoc/ignis`, etc or did you have something else in mind?

nonhermitian:
Yes, the stuff in apidocs needs to be cleaned up.  Basically the stopgap files need to be cleaned out.  Also, the main TOC file needs to be updated.  The latter step cannot be avoided when new modules are added to the elements.  I will get this working shortly.

chriseclectic:
Aers updated API docs has been merged into master now. Can those be copied into the top level index rather than nested under the `qiskit.providers.aer` subfolder?

nonhermitian:
Ok, once https://github.com/Qiskit/qiskit-aer/pull/443 is merged, this should build the docs from the master repos just fine.  Tried it with local installs and it worked good.

mtreinish:
Cool, thanks for updating. Just a couple questions/comments. The way the sphinx extension is written to clone the rst docs from each element right now it will not clone things if there is a pre-existing apidoc directory in the meta-repo. This was done so we could merge this without waiting for all the element updates.

The second thing is I think hard coding the filenames manually for each rst file in each element doesn't really scale, because we'll have to update the list in 2 places when adding a new rst file for each element. Could we just add a small rst file here with something like apidocs.rst with something like:

```
Qiskit API Reference:

.. toctree::
  :maxdepth: 2

Qiskit Terra <apidoc/qiskit>
Qiskit Aer <apidoc/qiskit_aer>
Qiskit Ignis <apidoc/qiskit_ignis>
Qiskit Aqua <apidoc/qiskit_aqua>
Qiskit IBMQ Provider <apidoc/qiskit_ibmq>
```

Then let each element own it's own section of the main toc?

nonhermitian:
A good point.  Let me play with that

nonhermitian:
@mtreinish that was a good idea and works locally.  I cannot seem to get `DOCS_FROM_MASTER=1 tox -edocs` to work with it though.  It looks like it is searching the `apidocs` dir before downloading the elements, so they are not getting indexed properly.

mtreinish:
I just tried it locally, I had to run `git clean -fdX` or `rm -rf docs/_build docs/stubs` to clean out any artifacts left from previous builds to get it to work. Then I also hit an issue where terra didn't update itself via https://github.com/Qiskit/qiskit/pull/551/commits/58420463649c26a97cc1f2a00a012f3cda0b9fa8 like the other elements did and still used 0.10.0 instead of master (which meant it didn't show all the docs, just visualization). I think there is a bug in that commit. Manually, installing it in the tox venv with (`.tox/docs/bin/pip install -U git+https://github.com/Qiskit/qiskit-terra`) and running it again worked.

mtreinish:
Oh, actually I spoke too soon even though the build succeeded and had all the module level docstrings from the `__init__` files as pages, none of the links to classes or functions from those pages worked.

docs/index.rst
```diff
@@ -39,7 +39,11 @@ Qiskit consists of four foundational elements:
   contributing_to_qiskit
   release_notes
   faq
-  API References <apidoc/qiskit>
+  Qiskit Terra API <apidoc/qiskit>
+  Qiskit Aer API <apidoc/qiskit_aer>
+  Qiskit Ignis API <apidoc/ignis/ignis>
```
SooluThomas:
```suggestion
  Qiskit Ignis API <apidoc/ignis>
```

docs/index.rst
```diff
@@ -39,7 +39,11 @@ Qiskit consists of four foundational elements:
...
+  Qiskit Aqua API <apidoc/aqua/aqua>
```
SooluThomas:
```suggestion
  Qiskit Aqua API <apidoc/qiskit_aqua>
```

docs/index.rst
```diff
@@ -39,7 +39,11 @@ Qiskit consists of four foundational elements:
...
+  Qiskit Aqua-Chemistry API <apidoc/chemistry/chemistry>
```
SooluThomas:
```suggestion
  Qiskit Aqua-Chemistry API <apidoc/chemistry>
```

docs/index.rst
```diff
@@ -39,7 +39,11 @@ Qiskit consists of four foundational elements:
...
   IBM Quantum Cloud Error Codes <errors/errors>
```
SooluThomas, line 47: Shouldn't qiskit-ibmq-provider API References be added too?

`Qiskit IBM Q Account <apidoc/ibmq>`

mtreinish:
With this latest version this is almost ready to merge. It moves ignis to remain in repo here until they're ready to push a new release.

The only issue left is that the autosummary stubs are not getting correctly generated and/or populated and  links are broken for elements in modules. This is the cause of these warnings: https://travis-ci.com/Qiskit/qiskit/jobs/286384546#L1403

mtreinish:
Ok, this is passing now and the stubs are included, @nonhermitian I think this is ready to merge

mtreinish (approved):

PR #557, opened 2019-09-12
add new localization contributor for DE localization

alfrisch:
adding my name to list of localization contributors.


alfrisch:
added issue for German localization in issue #558

derivation (approved):

PR #566, opened 2019-09-12
Changed the lang for sphinx make gettext

SooluThomas:
### Summary

This commit changes the logic to create `.po` files. 

Previously the `.po` files were generated from Japanese `.pot` files which is generated from `sphinx-build -b gettext` with `language=ja`. This will generate Japanese words on its own for some words like "Note", "Tip", "Warning" etc. So, the consequently created `.po` file will have the Japanese word in all languages. This commit sets the language to `en` to rectify this problem.

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -58,7 +58,6 @@ git config user.name "Qiskit Autodeploy"
 git config user.email "qiskit@qiskit.org"
 
 echo "git rm -rf for the translation po files"
-# git rm -rf --ignore-unmatch $DOC_DIR_2/$SOURCE_LANG/**/*.po # Remove old po files
 git rm -rf --ignore-unmatch $DOC_DIR_PO/$SOURCE_LANG/LC_MESSAGES/*.po \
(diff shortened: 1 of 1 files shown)


tools/deploy_documentation.sh
```diff
@@ -59,7 +58,7 @@ git config user.name "Qiskit Autodeploy"
 git config user.email "qiskit@qiskit.org"
 
 echo "git rm -rf for the translation po files"
-# git rm -rf --ignore-unmatch $DOC_DIR_2/$TRANSLATION_LANG/**/*.po # Remove old po files
+# git rm -rf --ignore-unmatch $DOC_DIR_2/$SOURCE_LANG/**/*.po # Remove old po files
```
mtreinish: We probably should just remove this line

```suggestion

```

mtreinish (dismissed):
I'm fine with this, but I'm wondering does it make sense to generate a `.pot` file to save some work on the translation. But I guess it's only a handful of extra strings to translate by hand.

mtreinish (approved):

PR #574, opened 2019-09-18
Add building docs at tags to deploy documentation

mtreinish:
### Summary

We want to start hosting docs at releases of qiskit so that users can
switch between them. This commit adds support for ensuring we've built
and hosted docs for each version (starting with 0.8.0). This patch adds
a new function build_old_versions and calls it before we add changes to
the target repo and push it. This function loops over all the tags and
checks if we've built the docs yet or not. If we haven't it will build
the docs at that version and add it to the target repo. This will make
the first build after this merges quite slow, but future builds will
have no overhead except for when we release which will be double length
to include the tag build in addition to the regular master build.


mtreinish:
Besides actually testing this (I just wrote the code and haven't tested it at all, so I assume it's broken) what this needs is theme support for switching between versions. Whether that be the read the docs theme or custom js/html adding a dropdown it doesn't matter. What is important is that we do no hard code a list of version in that dropdown (it will complicate the release process if we do). Once those 2 things are done I think this approach should work well.

mtreinish:
I've tested the main backfill loop locally in a test bucket here: http://qiskit-test.quantum-computing.ibm.com/documentation/stable/0.11.2/ and it looks like it works fine (it's still running) so I'm moving this out of WIP. Once that execution finishes I'll run a test with `$TRAVIS_TAG` set to verify the new stable builds work as expected.

SooluThomas (dismissed):

mtreinish:
Ok, the upload for all old tags worked fine, I also tested interuppting in the middle since we're going to hit timeouts on docs builds uploading old tag builds. I also tested with TRAVIS_TAG set to `0.13.1` to simulate what happens when we push a new tag. That also worked as expected and uploaded both the new tag build for that release: http://qiskit-test.quantum-computing.ibm.com/documentation/stable/0.13.1/ and also the translations: http://qiskit-test.quantum-computing.ibm.com/documentation/stable/0.13.1/locale/ja/

The only thing I noticed is that we'll have to update the drop-down widget for language because it is unconditionally going to documentation/locale/ instead of documentation/stable/$version/locale/. But as long as we fix that before the next release this should be good to go.

SooluThomas:
@mtreinish Thank you. Still working on the version switcher button.

3yakuya:
This PR could be a good place to introduce https://github.com/Qiskit/qiskit/issues/1024 (so that every function and/or property documentation contains info about the version where given element was introduced)

nonhermitian:
@mtreinish do we need this anymore?

mtreinish:
It needs to be updated and it is still partially blocked on having a small JS widget to do version discovery/selection but yeah it is still needed if we want to have documentation for old releases.

nonhermitian:
Oh yeah, I forgot about this.

mtreinish:
This has been superseded by #1261 (although I might steal the for loop from here to manually build the 0.24, 0.25, and 0.26 version of the documentation) for the starting history.

PR #628, opened 2019-10-10
Add benchmarks for converter functions

mtreinish:
### Summary

This commit adds benchmarks for the converter functions. The converter
functions are normally quick to execute but are commonly used as part of
other operations. For example circuit_to_dag and dag_to_circuit get
called under the covers during transpile() calls. Individually
benchmarking the performance of them is useful so we can isolate and
track their performance, even if it's normally only a small percentage
of a larger operation. The benchmarks here just use the large qasm
circuit example as a type of worst case example because it has >3k
gates. In the future we can add additional types of circuits such as
ones with many registers, etc to try and test additional worst case edge
cases.

### Details and comments

This was inspired by: https://github.com/Qiskit/qiskit-terra/pull/3176 which
should improve performance but there aren't any direct measurement of
that in the benchmark suite.


kdk (commented):
:+1: on the concept, and I agree we should be splitting our benchmarks into synthetic benchmarks (targeting smaller components, edge cases,  using our understanding of the implementation, ...) and application benchmarks (that closely cover what users commonly run).

Instead of `test_eoh_qasm`, can we build synthetic circuits from `circuit.random.random_circuit`? That way we can track with e.g. depth/width (`test_eoh_qasm` isn't great as either an application benchmark or a synthetic benchmark) . If there are other cases we'd like to cover (specifying a gate set so we can look at performance for gates that require many steps to reach the target basis), maybe we can add an issue in terra.

Otherwise, looks good.

mtreinish:
@kdk I switched this to use `random_circuit` in https://github.com/Qiskit/qiskit/pull/628/commits/95dee4171d30c606a60fd299614f426b1587ffb6 I also copied the random circuit function over so we can use it for the whole history of the converters module. As for width/depth numbers I tried to pick some values to give us wide coverage but not take too long. An iteration with `asv dev` took about 4min locally, depending on what you want to do we can remove some of the skips or change the parameters up as needed.

test/benchmarks/converters.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ConverterBenchmarks:
+    params = ([1, 2, 5, 8, 14, 20, 32, 53], [8, 128, 2048, 8192])
+    param_names = ['n_qubits', 'depth']
+    timeout = 600
+
+    def setup(self, n_qubits, depth):
+        seed = 42
+        if n_qubits >= 20:
```
kdk, line 35: Might be worth a comment explaining the exemption, so we clearer if it comes time to remove it.

mtreinish: Added in: https://github.com/Qiskit/qiskit/pull/628/commits/f4fa96b82f5e74084d8af3aeeb6f81f564e64fb5

kdk (approved):

PR #630, opened 2019-10-14
Add class's __init__() docstring too

mtreinish:
### Summary

Right now many classes document class constructors and their arguments
in the docstring for the `__init__()` methods of the class. However, by
default autodoc does not show these for the classes' documentation and
instead relies on only the class docstring. This commit sets the sphinx
autodoc config option [1] to use both the class and init docstring
together.

### Details and comments

[1] https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -277,6 +277,7 @@ epub_title = project
 # A list of files that should not be packed into the epub file.
 epub_exclude_files = ['search.html']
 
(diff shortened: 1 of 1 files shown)


SooluThomas (approved):
I agree! Thank you!

PR #635, opened 2019-10-16
Bump Meta

mtreinish:
Bump the meta repo version to include:

qiskit-terra==0.10.0
qiskit-aer==0.3.2
qiskit-aqua==0.6.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ matrix:
 language: python
 install: pip install -U tox
 script:
+  - pip install -U pip
   - tox
 
 notifications:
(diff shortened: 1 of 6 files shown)


kdk (dismissed):

mtreinish:
I've tagged this as `on hold` because I'd like to get an aer 0.3.2 out with: Qiskit/qiskit-aer#401 applied so we don't break a ton of people running with pip < 19.0. Also we need a new aqua release added to this that removes the terra cap so we can co-install aqua and terra 0.10.0.

mtreinish:
Removing on hold, this should be good to go now with the aer 0.3.2 release

nonhermitian (approved):

PR #641, opened 2019-10-21
Assemble circuits and transpiler peakmem benchmarks

kdk:
Adds a peakmem tracking for transpiler_level benchmarks, and a benchmark for assemble_circuits.

### DIFF
--- a/test/benchmarks/assembler.py
+++ b/test/benchmarks/assembler.py
@@ -15,7 +15,7 @@
 # pylint: disable=no-member,invalid-name,missing-docstring,no-name-in-module
 # pylint: disable=attribute-defined-outside-init,unsubscriptable-object
 
-from qiskit import assemble
+from qiskit.compiler import assemble
 
 from .utils import random_circuit
 
--- a/test/benchmarks/assembler.py
+++ b/test/benchmarks/assembler.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2019.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=no-member,invalid-name,missing-docstring,no-name-in-module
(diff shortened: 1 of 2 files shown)


test/benchmarks/assembler.py
```diff
@@ -0,0 +1,35 @@
...
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=no-member,invalid-name,missing-docstring,no-name-in-module
+# pylint: disable=attribute-defined-outside-init,unsubscriptable-object
+
+from qiskit import assemble
```
mtreinish: Depending on how far back we want to go on this it might make more sense to change this import to:

```suggestion
from qiskit.compiler import assemble
```

Since we added the top level import more recently than when the assemble function was added.

test/benchmarks/assembler.py
```diff
@@ -0,0 +1,35 @@
...
+
+from .utils import random_circuit
+
+
+class AssemblerBenchmarks:
+    params = ([1, 2, 5, 8],
```
mtreinish, line 24: Do we want to expand the number of qubits here at all. Just to get a feel for how this scales as we approach the largest sizes?

mtreinish (commented):
LGTM, a couple inline questions but nothing blocking

test/benchmarks/assembler.py
kdk: Yes and no. There are performance issues in `assemble` for larger `n_qubits` (e.g. we do some `List.index`s that should be dictionary look ups) but right now, far and away, performance scales with the number of gates. I limited `n_qubits` in the interest of keeping the benchmark runtime reasonable, but we can always expand it.

(As an aside, since we generate random circuits by depth and not by gate count, increasing `n_qubits` does dramatically increase the benchmark runtime but only because the test circuits contain more gates, so there's not a lot of interesting information there, at least right now. Maybe we should update `random_circuit` to accept either `depth` or `gate_count`.)

mtreinish: That's a good point, I figured it was for run time. Yeah most of the time is spent looping over the instructions in experiments.

If we wanted to add a gate_count parameter it'd be simple to do (my first implementation of random circuit which we didn't use in terra actually did it this way). We should do this before we adopt any benchmarks with it though because we'd have to regenerate any data.

mtreinish:
Ci Passed, not that we actually run benchmarks in CI (still waiting on the new asv release with the `--strict` flag). Travis issue seemed to report the job incorrectly so I went ahead and merged this.

PR #642, opened 2019-10-22
Fix visualization docs

nonhermitian:
### Summary
The visualization docs were being built twice due to the switch to module level doc strings in source.  Additionally `autosummary` was not doing its thing because it was not setup correctly.


docs/apidoc/qiskit.rst
```diff
@@ -25,4 +25,4 @@ API Reference
    tools/tools
    transpiler/transpiler
    validation
-   visualization/visualization
+   visualization
```
mtreinish, line 28: Should we remove the docs/apidoc/visualization directory too?

mtreinish (commented):
This LGTM. I noticed an issue with a stray `)` at the end of visualization on the side bar:

Screenshot_2019-10-24_08-50-49

But it looks like that's likely an issue with the terra source since it's there currently just one level lower (because of the duplication this fixes)

One inline question then I think this is good to go

docs/apidoc/qiskit.rst
nonhermitian: The dir should be removed and replaced with just a `visualization.rst` file with:

```

.. _qiskit-visualization:

.. automodule:: qiskit.visualization
   :no-members:
   :no-inherited-members:
   :no-special-members
```

mtreinish: oh, I just meant the PR still left the visualization dir intact. I've updated it here: https://github.com/Qiskit/qiskit/pull/642/commits/4195ad0ab8ed3d3779ee9d6300927aab3ac4071f

mtreinish (approved):

PR #653, opened 2019-10-25
Push directly to object store poBranch

mtreinish:
### Summary

This commit updates the documentation deploy job to stop pushing to the
Qiskit/qiskit.org website repository. There is no reason to do this
since the documentation static artifacts are easily generatable. Copying
these into another repo just needlessly increases the complexity of our
deployment jobs, and also makes the qiskit.org repo difficult to use
because it's history is bogged down with all these static artifacts,
especially multiple copies of binary files causing that repo to be
quite large. It also introduces another point of failure in the
deployment process. By pushing directly to cloud object storage for
hosting we make the deployment process faster and much more reliable.
When travis finishes executing on a deployment job we know the updated
documentation is being hosted.


SooluThomas:
@mtreinish Thank you!!

SooluThomas (approved):

PR #655, opened 2019-10-25
Parallelize sphinx builds for translated builds

mtreinish:
### Summary

Right now we're running the translated builds serially. This takes a lot
of time and will only get worse as we add more languages. To try and
mitigate this to a certain amount this removes the for loop in the
translated doc build with GNU parallel to run each translation build in
parallel. This should improve the speed of translation build jobs.


nonhermitian (approved):
Looks good.  Curious how translation_lang works given that the docs only show JA in the dropdown, but this is outside the scope of this PR.

SooluThomas:
> Looks good. Curious how translation_lang works given that the docs only show JA in the dropdown, but this is outside the scope of this PR.

@nonhermitian We haven't added dropdown for other languages since localization contributors are still working on translating the docs.

PR #661, opened 2019-11-01
Refactor master clone logic and set content_prefix to poBranch

SooluThomas:
### Summary

Edits made to poBranch

### DIFF
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -30,13 +30,15 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-cp -r docs_source/docs/. $SOURCE_DIR/docs/
+cp -r $SOURCE_DIR/docs/locale docs_source/docs
+rm -rf $SOURCE_DIR/docs/
+cp -r docs_source/docs/ $SOURCE_DIR/docs/
 
 pushd $SOURCE_DIR/docs
 
 # Make translated document
 sudo apt-get install parallel
-parallel sphinx-build -b html -D language={} . _build/html/locale/{} ::: $TRANSLATION_LANG
+parallel sphinx-build -b html -D content_prefix=documentation -D language={} . _build/html/locale/{} ::: $TRANSLATION_LANG
 
 popd
 
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -32,7 +32,7 @@ set -e
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
 cp -r $SOURCE_DIR/docs/locale docs_source/docs
 rm -rf $SOURCE_DIR/docs/
-cp -r docs_source/docs/ $SOURCE_DIR/docs/
+mv docs_source/docs/ $SOURCE_DIR/docs/
 
 pushd $SOURCE_DIR/docs
 
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -30,7 +30,7 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-rclone sync -v --exclude='locale/**' docs_source/docs docs/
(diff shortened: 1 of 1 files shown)


tools/deploy_documentation.sh
```diff
@@ -30,13 +30,15 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-cp -r docs_source/docs/. $SOURCE_DIR/docs/
+cp -r $SOURCE_DIR/docs/locale docs_source/docs
+rm -rf $SOURCE_DIR/docs/
+mv docs_source/docs/ $SOURCE_DIR/docs/
```
delapuente: I think this is `mv docs_source/docs $SOURCE_DIR`

tools/deploy_documentation.sh
```diff
@@ -30,13 +30,15 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-cp -r docs_source/docs/. $SOURCE_DIR/docs/
+cp -r $SOURCE_DIR/docs/locale docs_source/docs
```
delapuente: Adds a slash at the end of `docs`.

SooluThomas:
"on hold" because waiting for #662 to be completed and merged

delapuente (changes requested):

delapuente (approved):
I'm fine with this but let's wait for @mtreinish for another look.

tools/deploy_documentation.sh
```diff
@@ -30,13 +30,15 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-cp -r docs_source/docs/. $SOURCE_DIR/docs/
+cp -r $SOURCE_DIR/docs/locale docs_source/docs/
+rm -rf $SOURCE_DIR/docs/
+mv docs_source/docs/ $SOURCE_DIR
```
mtreinish: Is the point of this to do the equivalent of `rsync --exclude=locale --delete docs_source/docs/ $SOURCE_DIR/docs/` just using cp, rm, and mv? (ie to delete files that have been removed from master's docs) I'd honestly feel more comfortable using either rclone for this since we already have it installed for the upload and it mirrors the rsync mechanics. You can probably just replace these 3 lines with:

`rclone -v --exclude='locale/**' sync docs_source/docs docs/`

That way we can see when it deletes something or makes any other changes.

If you'd like to keep it like this I think we should at least add a comment here about the purpose of this section, because it few seconds to figure out the purpose of this.

tools/deploy_documentation.sh
```diff
@@ -30,13 +30,15 @@ set -e
...
 
 pushd $SOURCE_DIR/docs
 
 # Make translated document
 sudo apt-get install parallel
-parallel sphinx-build -b html -D language={} . _build/html/locale/{} ::: $TRANSLATION_LANG
+parallel sphinx-build -b html -D content_prefix=documentation -D language={} . _build/html/locale/{} ::: $TRANSLATION_LANG
```
mtreinish, line 39: Where does `content_prefix` come from? I can't find it in the sphinx documentation

delapuente: It is in our extension, to support the generation of the base urls for our translated and previous stable versions.

tools/deploy_documentation.sh
delapuente: Right! It is `rsync`. We can use `rclone`.

tools/deploy_documentation.sh
```diff
@@ -30,13 +30,13 @@ set -e
 
 # Clone the sources files and po files to $SOURCE_DIR/docs_source
 git clone --depth=1 $SOURCE_REPOSITORY docs_source
-cp -r docs_source/docs/. $SOURCE_DIR/docs/
+rclone sync -v --exclude='locale/**' docs_source/docs docs/
```
mtreinish: Rclone has implicit trailing slashes coming from rsync: https://rclone.org/commands/rclone_copy/

```suggestion
rclone sync -v --exclude='locale/**' docs_source/docs docs
```

mtreinish (approved):

PR #623, opened 2019-10-09
Fix typehints problem

SooluThomas:
### Summary

This PR partially fixes #569. 

### Details and comments

Extension `sphinx_autodoc_typehints` has been added to not show type hints in the signature but will be added to the docstring description. This can help in maintaining a consistency in the source code by just adding type hints in the signature.

#### BEFORE

image

#### AFTER

image

### TO-DO

- [ ] Figure out a way to show methods in a better way (**Edit** fixed in https://github.com/Qiskit/qiskit/pull/634)

image

PR #662, opened 2019-11-01
Change documentation theme to Read the Docs

### LINKED ISSUE #394: Documentation source code linebreaks

### LINKED ISSUE #467: Move to Read The Docs (RTD) theme
We should move to RTD to take advantage of the niceties that others are already using.

https://docs.readthedocs.io/en/stable/index.html

As discussed in #466

### LINKED ISSUE #490: doc is formatted incorrectly on phone
IMG_4887

### LINKED ISSUE #605: Increase width of source code text when viewed in the documentation
### **What is the current behavior?**

When viewing source code in the documentation, the text displayed is fixed at a narrow width, irrespective of width of the browser window and so requires horizontal scrolling.

### **What is the expected behavior?**

I would expect width of displayed source code to increase and decrease with width of browser window.

delapuente:
Fixes #467
Fixes #394
Fixes #605
Fixes #490
and also problems mentioned in #623

### Summary
Change the theme of documentation to Read the Docs (https://sphinx-rtd-theme.readthedocs.io/en/stable/)

### Details and comments
It also adds support for multiple languages, multiple versions (although it will be disabled right now), and include metrics.


docs/theme/versions.html
```diff
@@ -0,0 +1,31 @@
...
+  </div>
+  <script>
+    jQuery('.version').click((evt) => {
+      const hash = window.location.hash
+      const complete_url = evt.target.href + hash
+      window.location = complete_url
+      evt.preventDefault()
+    })
+  </script>
+</div>
```
SooluThomas:
```suggestion
</div>

```
Adding new line to the end of file

docs/versionutils.py
```diff
@@ -0,0 +1,177 @@
...
+    def run(self):
+        tags = _get_git_tags()
+        versions = self.get_versions(tags)
+        self.max_cols = len(self.headers)
+        self.col_widths = self.get_column_widths(self.max_cols)
+        table_node = self.build_table(versions)
+        title, messages = self.make_title()
+        if title:
+            table_node.insert(0, title)
+        return [table_node] + messages
```
SooluThomas:
```suggestion
        return [table_node] + messages
        
```
Adding new line to the end of file

mtreinish (changes requested):
The sphinx conf.py, extension changes, and the deploy_documentation and tox changes all look fine to me from a quick glance. But I built things locally and the theme configuration (I expect probably a css thing) don't look quite right to me. For example:

Screenshot_2019-11-04_17-32-59
Screenshot_2019-11-04_17-33-59

Everything is monospaced font except for code sections (which seems like the opposite of expectations).

Also one thing I like that @nonhermitian did for the terra docs builds (and something I've copied to aer and ignis) which is also using the RTD theme is having the qiskit logo in the upper left hand corner at the top of the side bar. For example:

Screenshot_2019-11-04_17-37-36

Maybe you should copy that from what terra has in it's local sphinx config.

delapuente:
You're right. This is happening in Safari. Let's find the cause and fix it. I'm delaying adding the Qiskit Logo (which personally I think it is a great idea) to the follow up #663

docs/versionutils.py
```diff
@@ -0,0 +1,178 @@
...
+    context['current_version'] = config.release
+    context['version_url'] = partial(_get_version_url, config)
+    context['version_label'] = _get_version_label(config)
+
+def _get_current_translation(config):
+    language = config.language or ''
+    return next(v for k, v in translations if k == language)
+
+def _get_translation_url(config, code, pagename):
+    base = '/locale/%s' % code if code else ''
```
SooluThomas, line 52: For pointing towards correct localized documentation, 
```suggestion
    base = '/documentation/locale/%s' % code if code else ''
```

delapuente: This is done by `_get_url` which is in charge of adding the content prefix.

SooluThomas: This has been taken care of in PR #661 [(change)](https://github.com/Qiskit/qiskit/pull/661/files#diff-b14ef4cdb78e1df5c71a1851ccd8f4efR39)

SooluThomas (dismissed):

docs/conf.py
```diff
@@ -155,64 +132,44 @@
...
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['theme/static/']
 
+html_context = {
+    'github_url': "https://github.com/Qiskit/qiskit",
+    'github_user': "Qiskit",
+    'github_repo': "qiskit",
+    'github_version': "master",
+    'conf_py_path': "/docs/",
+    'display_github': True,
```
mtreinish: I don't think we should do this actually. The edit on github button would be useful if the files contained details but since most of what we have is autodoc you end up at files like: https://github.com/Qiskit/qiskit/blob/master/docs/apidoc/execute.rst which don't really have the contents. Then after #551 merges the files won't even exist in Qiskit/qiskit anymore.

mtreinish (commented):
This looks better to me, the issues I outlined in the last review look like they're fixed now. I have one thing inline that we should change but I think this will be good to merge after that. I'm not a big fan of the color scheme I like what the terra docs do right now with the rtd theme:. But colors (and icons) are easy enough to iterate on post-merge and making the switch with all the widgets reimplemented is a big enough change on it's own.

mtreinish (commented):
This looks better to me, the issues I outlined in the last review look like they're fixed now. I have one thing inline that we should change but I think this will be good to merge after that. I'm not a big fan of the color scheme I like what the terra docs do right now with the rtd theme:. But colors (and icons) are easy enough to iterate on post-merge and making the switch with all the widgets reimplemented is a big enough change on it's own.

docs/theme/versions.html
```diff
@@ -0,0 +1,31 @@
...
+  <span class="rst-current-version" data-toggle="rst-current-version">
+    <span class="fa fa-book"> {{ _('You are viewing') }}</span>
+    v: {{ version_label }}
+    <span class="fa fa-caret-down"></span>
+  </span>
+  <div class="rst-other-versions">
+    {# Temporarily disabling version switching
+    <dl>
+      <dt>{{ _('Versions') }}</dt>
+      {% for version in version_list %}
```
mtreinish: This is actually a problem. Looking at how this is built at https://github.com/Qiskit/qiskit/pull/662/files#diff-95b277ed16282648f055393de56680cfR44 we can't do this statically with a build time variable. The build time list that we get from git is out of date the day we push a new tag. For example, when we add this when we push a 0.14.0 tag the statically hosted build we already uploaded for 0.13.0 won't have 0.14.0 in it's version_list variable. We need a way of getting the version info from a single source that we update independently of the individual doc build. For example we could have a static js file we continually update at `documentation/versions.js` that has the list or something, then make the theme search for that file off a relative url and display the version switcher from that data if that's present. As of right now we'll end up with a newer versions that aren't reachable from older versions which is less than ideal.

mtreinish (commented):
Sorry to keep finding one issue at a time, but I just realized that the version dropdown will have a pretty big problem as we move into the future. See inline

docs/theme/versions.html
delapuente: Yes, this is under the radar and @SooluThomas and I were discussing how to do it. Let's go and merge this one and we can start finding a solution for that and other version-related issues.

delapuente: A couple of solutions would be:
1. Use a daily build to check if there is some tag missing, and if so, trigger a deployment.
2. [Deploy on tag](https://docs.travis-ci.com/user/deployment-v2/conditional#tag).

Or even a combination of both.

delapuente:
Per offline conversations, getting the list of updated versions could be worked around. The real problem with static generation is that, if avoiding regenerating everything, every time a new version is added (i.e., building the latest one only), documentation for old versions would not get the list of new ones preventing the user from navigating to the latest version from a previous one.

delapuente:
@mtreinish @SooluThomas the latest commit remove everything related to versions. Translations can remain static until multi-version is added. A review would be appreciated.

SooluThomas (approved):

mtreinish (approved):
I think this LGTM. I tested a local build and everything looks good. I still think there are some small style things to fix, but that's easy to do in a follow-up. Thanks for updating it to address my earlier concerns.

mtreinish:
I held off on merging this because of #664. Without a fix for that the documentation will become a blackhole from qiskit.org once people click on it. We should at least have a solution identified for that so we can push a quick follow up to this and minimize the amount of time things are in that state

mtreinish:
> I think this LGTM. I tested a local build and everything looks good. I still think there are some small style things to fix, but that's easy to do in a follow-up.

I wrote a patch for this on top of this branch here: https://github.com/mtreinish/qiskit/commit/d6bb3bd08e6861585443b920e82e8117ff23efad I'll push a PR with it as soon as we merge this.

delapuente:
Thank you very much @SooluThomas and @mtreinish

PR #674, opened 2019-11-08
Updating logo

SooluThomas:
### Summary

Update the logo.

### Details and comments

Rather than having just a Bloch sphere, adding "Qiskit" to the image below the logo.

### Screenshot

image


mtreinish (changes requested):
So I'm fine with putting the text on the logo image, but looking at this version of the logo image I don't think it looks better. At least to my eyes.

For example, in this zoomed in screenshot:

Screenshot_2019-11-08_15-01-19

the logo adds a 3rd different font for "qiskit" in just the upper left hand corner.

The text doesn't also look quite centered, it's more to the left of center. It's especially noticeable when looked at with the search bar:

Screenshot_2019-11-08_15-04-28

I think having no text like now is preferable until we address these issues in the logo with text.

SooluThomas:
The logo is in `.svg` and looks like this is happening only in Safari. I'll push a commit with `.png`.
>The text doesn't also look quite centered, it's more to the left of center. It's especially noticeable when looked at with the search bar:

The PNG addresses the above
>I think having no text like now is preferable until we address these issues in the logo with text.

nonhermitian:
I think the png solves this issue.  Note that setting this line to False will make a little Qiskit link as well under the logo:

 https://github.com/Qiskit/qiskit/blob/8a80dc33026747cf0f149a4d46d1c883b11ae710/docs/conf.py#L144

mtreinish:
> I think the png solves this issue. Note that setting this line to False will make a little Qiskit link as well under the logo:
> 
> https://github.com/Qiskit/qiskit/blob/8a80dc33026747cf0f149a4d46d1c883b11ae710/docs/conf.py#L144

I would be fine with this approach, I think it's preferable to trying to add the text ourselves to the image. Especially because it will reflect whatever changes we make to the theme font-wise without having to edit an image.

SooluThomas:
I'm also in favor of this suggestion for now, until we have a logo with text (aesthetic reasons). 

> I think the png solves this issue. Note that setting this line to False will make a little Qiskit link as well under the logo:
> 
> https://github.com/Qiskit/qiskit/blob/8a80dc33026747cf0f149a4d46d1c883b11ae710/docs/conf.py#L144

image

nonhermitian:
Closing this as it seems stale.

PR #683, opened 2019-11-12
Use IBM Plex in headers and other elements

delapuente:
### Summary
Use IBM Plex in headers and other HTML elements.


SooluThomas (approved):
Thank you @delapuente

PR #687, opened 2019-11-13
Add travis job that runs a docs build

mtreinish:
### Summary

Right now we are only running a linter on the docs, we never try to
actually build the documentation before we merge a patch. This
potentially exposes us to introducing a chance that doesn't work. It's
unlikely (and I don't think it's actually ever happend) because doc8
should catch most of the sphinx formatting errors, but it's not 100%.
To make sure this doesn't happen this commit adds a ci job that runs a
docs build to make sure we can compile the documentation at least.

### Details and comments


kdk (approved):

taalexander (approved):
Lgtm

PR #691, opened 2019-11-17
Added better navigation to different elements under Installation from source section.

### LINKED ISSUE #690: Cleanup of documentation inside contributing_to_qiskit.rst.
Yes
- Which files need to be updated?
 qiskit/docs/contributing_to_qiskit.rst 
- What content needs to be updated or replaced? (You can copy and paste)

Installing from source section doesn't provide better navigation to these elements installation.

    qiskit-terra
    qiskit-aer
    qiskit-ignis
    qiskit-aqua
    qiskit-ibmq-provider (if wanting to connect to the IBM Q devices or online simulator)

- What new content needs to be added?

I think it will be better if these has navigation to its own sections like this.

Screenshot 2019-11-17 at 4 54 05 PM

melvin2016:
Fixes #690
### Summary
Added better navigation to different elements under Installation from Source section.

Screenshot 2019-11-17 at 4 54 05 PM


derivation (changes requested):
Thank you for contributing. Because the docs get translated into multiple languages, it's best to use the `:ref:` technique rather than rely on the names of sections, which will change from language to language. Here's an example of the `:ref:` syntax:

`.. _my-reference-label:`

`Section to cross-reference`
`--------------------------`

`This is the text of the section.`

``It refers to the section itself, see :ref:`my-reference-label`.``

melvin2016:
Ok i will make the changes.

melvin2016:
> Thank you for contributing. Because the docs get translated into multiple languages, it's best to use the `:ref:` technique rather than rely on the names of sections, which will change from language to language. Here's an example of the `:ref:` syntax:
> 
> `.. _my-reference-label:`
> 
> `Section to cross-reference`
> `--------------------------`
> 
> `This is the text of the section.`
> 
> `` It refers to the section itself, see :ref:`my-reference-label`. ``

Thank You for pointing out the mistake. I had made the changes.
@derivation

derivation (changes requested):
Thank you. Could you prepend the word 'install' to the :ref: labels (e.g. _install-qiskit-terra) to increase the likelihood that the labels will be unique across the entire doc? The :ref: references work across all of the doc files, not just within a single file.

docs/contributing_to_qiskit.rst
```diff
@@ -190,11 +190,11 @@ Installing elements from source requires the following order of installation to
...
-#. qiskit-terra
-#. qiskit-aer
-#. qiskit-ignis
-#. qiskit-aqua
-#. qiskit-ibmq-provider (if wanting to connect to the IBM Q devices or online
+#. qiskit-terra - :ref:`qiskit-terra`
+#. qiskit-aer - :ref:`qiskit-aer`
+#. qiskit-ignis - :ref:`qiskit-ignis`
+#. qiskit-aqua - :ref:`qiskit-aqua`
+#. qiskit-ibmq-provider - :ref:`qiskit-ibmq-provider` (if wanting to connect to the IBM Q devices or online
```
SooluThomas: I'd also suggest using 
```
#. :ref:`install-qiskit-terra <qiskit-terra>`
#. :ref:`install-qiskit-aer <qiskit-aer>`
#. :ref:`install-qiskit-ignis <qiskit-ignis>`
#. :ref:`install-qiskit-aqua <qiskit-aqua>`
#. :ref:`install-qiskit-ibmq-provider <qiskit-ibmq-provider>` 
```
so that the output generates the same old thing with hyperlinks.

melvin2016: Yeah this seems more friendly. Adding this.
I think the order is vice-versa 
``` #. :ref:`qiskit-terra <install-qiskit-terra>` ```
 instead of 
``` #. :ref:`install-qiskit-terra <qiskit-terra>` ```.

melvin2016:
@SooluThomas @derivation pushed a commit with proposed changes.

derivation (dismissed):

derivation (approved):

derivation:
Thanks for your help, @melvin2016

PR #704, opened 2019-11-21
Add docs/stubs to .gitignore

SooluThomas:
### Summary

This commit avoids accidental adding of unwanted `docs/stubs` files.


derivation (approved):
Thank you. This will avoid accidental pushes of doc generated from the various Qiskit elements.

PR #715, opened 2019-12-04
Travis CI: Add Python 3.8 to the testing

cclauss:
The sudo tag is now deprecated and Linux and Xenial are defaults.

### Details and comments

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,23 +15,19 @@ matrix:
       env: TOXENV=py35
     - python: "3.6"
       env: TOXENV=py36
-    - os: linux
-      dist: xenial
-      python: "3.7"
+    - python: "3.7"
       env: TOXENV=py37
-      sudo: true
-    - python: "3.5"
+    - python: "3.8"
+      env: TOXENV=py38
+    - python: "3.8"
       env: TOXENV=lint
     - if: branch = master AND type = push
-      os: linux
-      dist: xenial
-      python: "3.7"
-      sudo: true
+      python: "3.8"
       script: tools/deploy_documentation.sh
       git:
         depth: false
     - if: tag IS present
-      python: "3.6"
+      python: "3.8"
       env:
         - TWINE_USERNAME=qiskit
       install: pip install -U twine
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.1
-envlist = py35, py36, py37, pep8
+envlist = py35, py36, py37, py38, pep8
 skipsdist = True
 
 [testenv]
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -15,23 +15,19 @@ matrix:
...
-      dist: xenial
-      python: "3.7"
-      sudo: true
+      python: "3.8"
       script: tools/deploy_documentation.sh
       git:
         depth: false
     - if: tag IS present
-      python: "3.6"
+      python: "3.8"
```
mtreinish, line 30: Why change the version we build the sdist with?

mtreinish (commented):
The python 3.8 side of this PR is a bit premature and was something we were going to add eventually. But it won't work yet because the elements that are used to build the meta package do not support python 3.8 yet. See: https://github.com/Qiskit/qiskit-terra/pull/3268 https://github.com/Qiskit/qiskit-ignis/pull/298 https://github.com/Qiskit/qiskit-aer/pull/428 https://github.com/Qiskit/qiskit-aqua/pull/722 https://github.com/Qiskit/qiskit-ibmq-provider/pull/445

As for the lint job changes, quite frankly adding additional python style and rule checks to the meta repository is a waste of time. The repo contains basically 0 code outside tests and benchmarks so the extra checking really doesn't add much value.

That being said I feel like we have the needle for style rules too far already and I'm worried about adding more checks. I am generally a fan of flake8 and use it as the sole tool for most of my personal projects. But when we combine it with pylint we'll end up in a situation where our style enforcement is overly pedantic and we spend a great deal of time trying to thread the needle between all the rules enforced by the different tools  (I have an issue open about this on terra already: https://github.com/Qiskit/qiskit-terra/issues/1179 ). Another thing worth considering though is flake8 wraps pycodestyle already and we probably could just replace pycodestyle with flake8 to get the expanded set of checks  (although I'm still hesitant about adding more rule checks).

cclauss:
Closing.  As it says in the first line of the commit message, these are _not_ style rules.

PR #751, opened 2019-12-22
fix typo in files under docs and test

kiszk:
### Summary

Fix typo and spelling in rst files under `docs` directory and in python file under `test` directory


test/benchmarks/passes.py
```diff
@@ -176,7 +176,7 @@ def peakmem_num_tensor_factors(self, _, __):
     def time_resource_optimization(self, _, __):
         ResourceEstimation().run(self.dag)
 
-    def peakmem_resoure_optimization(self, _, __):
+    def peakmem_resource_optimization(self, _, __):
```
mtreinish: While I agree this is a typo changing this would invalidate the history of results for this benchmark hosted at:

https://qiskit.github.io/qiskit/#passes.PassBenchmarks.peakmem_resoure_optimization

I think weighing a typo correction vs having to lose the historical data we should just leave the typo.

mtreinish (changes requested):
Looking solely at the `test/benchmarks/passes.py` file this lgtm. Catching the 2 typos on `peakmem` adds 2 benchmarks that haven't been running (`peakmem_` is a prefix for the benchmark suite telling it how to run the benchmark). However as noted inline one typo I think we have to live with for the time being otherwise we'll invalidate the historical data already built for that benchmark.

test/benchmarks/passes.py
kiszk: @mtreinish Thank you for letting me know the important information. I did not know it. I totally agree with reverting changes in `passes.py`.

mtreinish: @kiszk thanks for updating, but you reverted a bit too much in the latest revision. Doing the `s/peakemem/peakmem/g` replacement was the correct thing to do.

The only thing which I was saying we shouldn't change was `s/resoure_optimization/resource_optimization/g` because of the historical tracking. If you could add back the original `peakemem` fixes then I'll be good with this. Thanks.

mtreinish (dismissed):
Thanks for updating, this LGTM

jaygambetta (approved):

mtreinish (approved):

PR #768, opened 2020-01-09
Error mitigation RFC

eggerdj:
## Summary
This PR contains an RFC discussing error mitigation and how to include it in Qiskit.

### Details and comments
Details are in  rfcs/0000-error-mitigation.md


rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+**The limitations of this method are**:
+- It requires a lot of knowledge from the user.
+- It requires that the user run many calibration jobs in addition to his quantum circuit.
+
+## Detailed Design
+Here we focus on the implementation details of the Backend constrained error mitigation.
+The backend will have a set of calibrated gates with different stretch factors that will be made available to Qiskit through the config file.
+
+```
+ in: config = backend.configuration().to_dict()
```
lcapelluto: Should avoid this, and encourage:
```
config = backend.configuration()
print(config.stretch_factors)
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+ in: print(config['stretch_factors'])
+out: [1.0, 1.1, 1.25, 1.5]
+```
+
+To execute a quantum circuit, the user would do
+```
+execute(cirq, backend, ..., error_mitigation=True)
```
lcapelluto: rather than a bool, we should use method names so that multiple techniques can be supported, e.g. `error_mitigation='method_X'`

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+- It requires that the user run many calibration jobs in addition to his quantum circuit.
+
+## Detailed Design
+Here we focus on the implementation details of the Backend constrained error mitigation.
+The backend will have a set of calibrated gates with different stretch factors that will be made available to Qiskit through the config file.
+
+```
+ in: config = backend.configuration().to_dict()
+ in: print(config['stretch_factors'])
+out: [1.0, 1.1, 1.25, 1.5]
```
lcapelluto: What if for example `CX12` supports different stretch factors than `CX23`? Or what if one is missing? Does this apply to 2Q gates only?

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+execute(cirq, backend, ..., error_mitigation=True, stretch_factors=[1.0, 1.25, 1.5])
+```
+to use only a subset of the stretch factors.
+The changes needed in Qiskit to implement error mitigation would require a pre-prossessing step in assemble to convert the quantum circuit (or list of quantum circuits) into a list of schedules, using the scheduler, that include the stretched pulses.
+This also implies that error mitigation will only be a meaningful option for quantum circuits and not for schedules.
+For instance, code like the following would be required in `assemble` 
+```
+if error_mitigation and all(isinstance(exp, QuantumCircuit) for exp in experiments):
+    schedule_config['stretch_factors'] = stretch_factors
+    error_mitigation_schedules = schedule_circuit_error_mitigation(experiments, schedul_config, method)
```
lcapelluto: `schedule` has a method arg, this could be something like
```suggestion
    error_mitigation_schedules = schedule(experiments, schedule_config, method='stretch_factor_error_mitigation')
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+    
+    return assemble_schedules(schedules=error_mitigation_schedules, qobj_id=qobj_id,
```
lcapelluto: something funky might happen elsewhere to user code (or internal to qiskit maybe, too) when the number of jobs executed is not equal to the number of circuits given, since this is often assumed

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+    PulseQobjInstruction(ch='d1', name='CR90p_d1_4334', t0=160), 
+    PulseQobjInstruction(ch='u0', name='CR90p_u0_0e7b', t0=160), 
+    PulseQobjInstruction(ch='d0', name='Xp_d0_e1b0', t0=672), 
+    PulseQobjInstruction(ch='d1', name='CR90m_d1_f19b', t0=832), 
+    PulseQobjInstruction(ch='u0', name='CR90m_u0_0cdb', t0=832)
+])
+```
+The definition of this gate could be extended to
+```
+Command(name='cx', qubits=[0, 1], stretch_factor=1.1, sequence=[
```
lcapelluto: maybe we can do something more flexible, like `meta_data={'stretch_factor'=1.1}`. Probably not exactly this, because `Command` is for serializing, and I don't think they support dicts

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+    PulseQobjInstruction(ch='d0', name='fc', phase=1.5707963267948966, t0=0),
+    ...
+])
+```
+to include information on the stretch factor which can be used by `translate_gates_to_pulse_defs` to select the gates with a stretch factor corresponding to `schedule_config['stretch_factor']` (this is a float and not a list).
+This would then allow the creation of schedules for error mitigation that can be assembled using `assemble_schedule` and run on the backend.
+
+Here are some additional considerations:
+- Currently, the name of a scheduled circuit is the same as the circuit. We will also need to distinguish the schedules with different stretch factors, for instance, by including the stretch factor in the name of the circuit. E.g. `sched = Schedule(name=circuit.name + 'c=%d'.format(schedule_config['stretch_factor']))`.
```
lcapelluto: Maybe `Schedule`s should also have metadata. I don't really like the idea of using `name` to hold data that will be used programmatically elsewhere

lcapelluto: You are still right that the names will have to be different (because unfortunately `name` might already be used to grab one schedule's results from a list of a job's results)

rfcs/0000-error-mitigation.md
eggerdj:
```suggestion
 in: config = backend.configuration()
 in: print(config.stretch_factors)
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+- It requires a lot of knowledge from the user.
+- It requires that the user run many calibration jobs in addition to his quantum circuit.
+
+## Detailed Design
+Here we focus on the implementation details of the Backend constrained error mitigation.
+The backend will have a set of calibrated gates with different stretch factors that will be made available to Qiskit through the config file.
+
+```
+ in: config = backend.configuration().to_dict()
+ in: print(config['stretch_factors'])
```
eggerdj:
```suggestion
```

rfcs/0000-error-mitigation.md
eggerdj: good point. This will make it easier to extend.

rfcs/0000-error-mitigation.md
eggerdj: This will apply to single-qubit and two-qubit gates. Ideally, if a stretch factor is present in `config` then all gates should be calibrated with this stretch factor. If one or several gates are not calibrated for a stretch factor then either:
a) do not include this stretch factor in `config`
b) default to the closest stretch factor, but I think this is cheating and incorrect.
The point you raise here shows that, in this implementation, it is the backend's responsibility to ensure that all gates are calibrated with at least three stretch factors.

rfcs/0000-error-mitigation.md
eggerdj: Do you know where in qiskit this might be an issue?

rfcs/0000-error-mitigation.md
eggerdj: I agree with you on both points. Each schedule will still need a unique name but putting the stretch factor in the schedule name is not particularly appealing. Lets come up with something else.

rfcs/0000-error-mitigation.md
eggerdj: Perhaps we could introduce the `stretch_factor` property in `Command`? However, this does not make things flexible. One extra point is that, we should avoid cluttering the data that is transported. For instance, if backends do not implement error mitigation it does not seem reasonable to have all commands send over a `'stretch_factor'=1.0`. Thus, by default we assume that if stretch factor is not present then it is 1.0. The converter would do something along the lines:
```
command_dict = {
    ...
}

if instruction.stretch_factor != 1.0:
    command_dict['stretch_factor'] = instruction.stretch_factor
```
And all commands would by default have stretch factor = 1.0.

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+
+```
+ in: config = backend.configuration()
+ in: print(config.stretch_factors)
+out: [1.0, 1.1, 1.25, 1.5]
+```
+
+To execute a quantum circuit, the user would do
+```
+execute(cirq, backend, ..., error_mitigation=True)
```
eggerdj:
```suggestion
execute(cirq, backend, ..., error_mitigation='richardson')
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+```
+to use all the stretch factors or 
+```
+execute(cirq, backend, ..., error_mitigation=True, stretch_factors=[1.0, 1.25, 1.5])
```
eggerdj:
```suggestion
execute(cirq, backend, ..., error_mitigation='richardson', stretch_factors=[1.0, 1.25, 1.5])
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,166 @@
...
+```
+to use only a subset of the stretch factors.
```
eggerdj:
```suggestion
to use only a subset of the stretch factors.
Here, `error_mitigation` specifies the error mitigation method to use.
This value would be default be `None` when error mitigation is not used.
```

rfcs/0000-error-mitigation.md
lcapelluto: I think this is an example, sort of pseudocode:
```
job = execute(circuits)  # circuits: List[QuantumCircuit]
result = job.get_results()
result_for_ith_circuit = result.get_counts(i)
```
I always see something like this on the user side for retrieving their data

lcapelluto: specifically, I think the Result object or however results are returned could get hairy. I find the results object already pretty confusing, so I'm not sure what changes would have to be made or what issues to expect

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,168 @@
...
+execute(cirq, backend, ..., error_mitigation='richardson')
+```
+to use all the stretch factors or 
+```
+execute(cirq, backend, ..., error_mitigation='richardson', stretch_factors=[1.0, 1.25, 1.5])
+```
+to use only a subset of the stretch factors.
+Here, `error_mitigation` specifies the error mitigation method to use.
+This value would be default be `None` when error mitigation is not used.
+The changes needed in Qiskit to implement error mitigation would require a pre-prossessing step in assemble to convert the quantum circuit (or list of quantum circuits) into a list of schedules, using the scheduler, that include the stretched pulses.
```
ajavadia: Why does this need to become a list of schedules? To me, the advantage of the backend-constrained method is precisely that we can use some pre-calibrated gates, i.e. we just send a circuit. There is no need to create pulse schedules if the backend provides those stretched gates.

rfcs/0000-error-mitigation.md
ajavadia: The backend configuration schemas are written in a way to allow maximum flexibility, in case of heterogeneity in the device (different calibrated gate on different qubits, etc.). 

So there is actually a place in the current schemas to include this, which would be: https://github.com/Qiskit/qiskit-terra/blob/e34a259decc778ff1395e8665e55955356022b2b/qiskit/schemas/backend_configuration_schema.json#L255

You can then include different stretch factors for different gates. 

I agree that if you want to do mitigation, you need similar stretch factors on your used gates. But making the strict requirement that a device cannot be deployed unless all gates are calibrated with 4 stretches will be hard to adhere to. The user/compiler can see what stretch factors are implemented and make decisions accordingly.

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,168 @@
...
+out: [1.0, 1.1, 1.25, 1.5]
+```
+
+To execute a quantum circuit, the user would do
+```
+execute(cirq, backend, ..., error_mitigation='richardson')
+```
+to use all the stretch factors or 
+```
+execute(cirq, backend, ..., error_mitigation='richardson', stretch_factors=[1.0, 1.25, 1.5])
```
ajavadia:
```suggestion
execute(circ, backend, ..., error_mitigation='richardson', stretch_factors=[1.0, 1.25, 1.5])
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+ in: config = backend.configuration()
+ in: print(config.stretch_factors)
+out: [1.0, 1.1, 1.25, 1.5]
+```
+
+To execute a quantum circuit, the user would do
+```
+execute(cirq, backend, ..., error_mitigation='richardson')
```
ajavadia:
```suggestion
execute(circ, backend, ..., error_mitigation='richardson')
```

ajavadia: Can you also discuss how the different circuits that are created "under the hood" will be handled? For example currently when we `execute([circ1, circ2], backend)`, there will be a `qobj` created that gets sent over the wire. There is a `qobj.experiments` field which in this case will contain instructions associated with `circ1` and `circ2`.
Here, your 2 circuits will get converted to 8 experiments. The RFC should discuss how those will be names, tracked in the Qobj and Result, etc. The Result as is now will contain 8 counts, but obviously we want to richardson-extrapolate the inner ones to get 2 counts. I'm thinking this actually can be done with Ignis (it fits the circuit generation+fitter+filter model of Ignis).

ajavadia (commented):
I left some comments. I like the backend-calibrated gates. I think similar to this, we can do easier readout error mitigation if the backend provides some extra data. But that's for another RFC.

rfcs/0000-error-mitigation.md
eggerdj: Thanks @ajavadia. The example above is just to illustrate. The backend should not be required to support exactly 4 stretch factors. The backend may well chose not support error mitigation in which case there will only be one set of default pulses (corresponding to c=1.0). Backends that implement error mitigation may do so with as many stretch factors as desired. I suggest 4 above as example since this is what was used in the original error mitigation paper and seems reasonable.

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,168 @@
...
+- This allows error mitigation to be applied on single-qubit and two-qubit gates.
+- The stretch factors may be chosen as a function of the depth of the quantum circuit.
+
+**The limitations of this method are**:
+- It requires a lot of knowledge from the user.
+- It requires that the user run many calibration jobs in addition to his quantum circuit.
+
+## Detailed Design
+Here we focus on the implementation details of the Backend constrained error mitigation.
+The backend will have a set of calibrated gates with different stretch factors that will be made available to Qiskit through the config file.
```
eggerdj:
```suggestion
The backend will have a set of calibrated gates with different stretch factors that will be made available to Qiskit through the config file.
For example, a backend that calibrated its default set of gates with four different stretch factors would have
```

rfcs/0000-error-mitigation.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+ in: config = backend.configuration()
+ in: print(config.stretch_factors)
+out: [1.0, 1.1, 1.25, 1.5]
+```
+
```
eggerdj, line 138:
```suggestion
A backend may implement as many stretch factors as is deemed reasonable by those who maintain the backend.
A backend may also have the choice to not implement error mitigation at all.
The `gateconfig` schema (see `qiskit/schemas/backend_configuration_schema.json`) will be updated to support the stretch factor as follows

```

rfcs/0000-error-mitigation.md
eggerdj: Right, I think the point is the following. Do we explicitly send, e.g., X times the circuit to the backend. Or do we send 1 circuit to the backend and tell the backend to run error mitigation on it with X stretch factors?

rfcs/0000-error-mitigation.md
eggerdj: Yes let's use Ignis for this. I'll also think about how the qobj handles the error mitigation.

rfcs/0000-error-mitigation.md
ajavadia: I'm not worried about the number of stretch factors. That can be any number.

My issue is with the suggested place for including these stretch factors in the configuration. `config['stretch_factors']` is a top-level attribute of the config. This would imply that *every* gate has the same stretch factors calibrated. What if we only calibrate stretch factors on gates involving qubits 0-4 in a 20-qubit device? 

That's why I suggested the above location to make the `configuration` schema change, which keeps the flexibility of defining and reporting stretch factors on a per-gate basis, rather than globally.

rfcs/0000-error-mitigation.md
ajavadia: Backends don't currently have capabilities for classical processing of this kind (e.g. richardson extrapolation on server side). This has to happen on client side in the near term. 

My only issue is with the phrase "list of schedules". This implies we are running OpenPulse experiments. However, in reality we are running QASM experiments. (where we use newly calibrated gates in our QASM). So it should be "list of circuits".

eggerdj: Solved. I updated this in the recent PRs. We can do the extrapolation with Ignis and the backends will received a list of circuits of length nbr stretch factors times nbr of original circuits.

rfcs/0000-error-mitigation.md
eggerdj: Ah right, I see what you mean @ajavadia. Please see the newest version (i.e. f08999d and onward) this should address this issue.

ajavadia:
@lcapelluto please make new comments on https://github.com/Qiskit/rfcs/pull/3. I'll also do a final review there.

PR #787, opened 2020-01-16
Add contributors not associated with any commit

Cryoris:
### Summary

Add a whitelist for contributors that have no commits associated with their name.

### Details and comments

Some contributors to Qiskit have no commits associated with their name, but still contibuted to Qiskit, e.g. by 
* contributing to previous Qiskit versions in older repos, or
* having co-authored code. 
They should also co but should also appear in the bib-file.
Therefore this PR adds a `whitelisted_authors` list where these cases can be manually added.

This closes Qiskit/qiskit-aqua#766.


mtreinish (approved):
LGTM, thanks. I agree we need to do this for past contributions where we weren't giving credit to co-authors with a co-authored-by line in the commit message.

As an aside for future PR maybe not changing whitespace as part of unrelated lines would make this a bit easier to review. It was a bit more difficult to figure out the exact path being added here because of the unrelated changes.

Cryoris:
Thanks!

Yes, I'll make sure to prevent that next time (or to prevent the editor of doing that).

PR #807, opened 2020-02-12
Print realtime upload progress from rclone

mtreinish:
### Summary

For particularly big docs diffs the upload to object store is taking
longer than 10mins. Travis will terminate a job that is taking longer
than 10mins without output. To avoid this issue this commit adds the
--progress flag to the rclone sync command to add some output as it
uploads.

### Details and comments


SooluThomas (approved):
Thank you!

PR #810, opened 2020-02-16
Update install.rst

jwoehr:
Information on VC++ runtime was out of date

### Summary
Info on install under Windows + Python 3.7 is out of date


jwoehr:
For that matter, perhaps we should not mention WIndows 7 ... if you've been following the news, that has rapidly turned into a steaming heap since EOL.

jwoehr:
image

jaygambetta (dismissed):

mtreinish (commented):
Sorry for the delay in reviewing this. But I'm not sure that this is accurate, you shouldn't need to install visual studio to install qiskit. You will definitely need it to build terra and aer from source (or from sdist), but if you're installing from the precompiled wheels it should work without needing anything extra. It has been a while since I tested it in my windows vm, but last I remember just installing python and doing `pip install -U qiskit` worked without anything extra. Was this for the source installation or the general install?

nonhermitian:
A fresh pip install on Windows 10 using Py37 from Anaconda did not require having VS.

mtreinish:
Yeah, I just tested this even outside of anaconda on a fresh install of python from python.org nothing extra was needed to be installed to pip install qiskit. I would just remove that entire block

jaygambetta:
ok removed it

mtreinish (approved):

PR #817, opened 2020-02-18
Sync setup.py to poBranch too

mtreinish:
### Summary

As part of the migration to having apidocs live solely in the individual
elements was that the sphinx builds now rely on having up to date
version info in the setup.py for the meta package. However, on poBranch
where we build the translated version of the docs we were never updating
setup.py. This currently results in broken builds because the sphinx
plugin for pulling the docs source from the released elements can't
parse the terra version info, but in the future it would also result in
apidocs that were out of sync for translated docs. This commit fixes
these by also syncing setup.py from master to poBranch as part of deploy
documentation.


SooluThomas (approved):
Thank you 😊

PR #822, opened 2020-02-18
Bump numpy on poBranch

mtreinish:
### Summary

The translated docs builds are failing because of Qiskit/qiskit-aer#592
which is caused by using a version of numpy incompatible with the pulse
simulator. This has been fixed in aer on master and the stable/0.4.0
branch pending a 0.4.1 release. But, until that is ready we need to
manual ensure we're using a newer numpy version to avoid this issue in
the docs builds.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
         - twine upload dist/qiskit*tar.gz
 
 language: python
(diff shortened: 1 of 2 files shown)


SooluThomas (approved):
thank you!

PR #823, opened 2020-02-19
Add provider 0.4.6 release note

jyu00:
### Summary
Add release note for qiskit-ibmq-provider 0.4.6 (Qiskit 0.15.0)

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -740,48 +740,50 @@ IBM Q Provider 0.4.6
 Added
 -----
 
-- ``IBMQJob`` now has a new method ``wait_for_final_state()`` that blocks
-  until the job finishes. One of its parameters is a callback function
-  that it will invoke after every query to provide feedback. (\#529)
-- ``IBMQBackend`` now has a new method ``active_jobs()``. The method returns the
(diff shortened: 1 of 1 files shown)


mtreinish (commented):
I'm confused why there are so many new features in a patch fix. Shouldn't this have been 0.5.0 per the stable branch and versioning policy? https://qiskit.org/documentation/contributing_to_qiskit.html#stable-branch-policy and https://qiskit.org/documentation/development_strategy.html#versioning

That being said could you do 2 things before we merge this, switch the function and method references to links using the sphinx syntax and the pr numbers to links? This isn't really release documentation it's just the changelog. I get that this is what's there for aer, but that was supposed to be a temporary thing while @chriseclectic was writing real release documentation.

jyu00:
> Shouldn't this have been 0.5.0...

In retrospect it probably should have been. The changes were all very minor so I just went with a patch release. 

> That being said could you do 2 things before we merge this, switch the function and method references to links using the sphinx syntax and the pr numbers to links? 

I cleaned things up a bit and added the reference links. I removed the PR numbers though since they were not used in the terra section.

mtreinish (approved):

PR #828, opened 2020-02-19
Split translation doc publish jobs for each language

mtreinish:
### Summary

Right now we're stuck in a spot where the docs builds can't be run in
parallel locally and doing 2 language builds serially exceeds the
maximum time allowed for a travis job. This commit tries to workaround
this by splitting the jobs up so we have an individual job for each
language. This should avoid the parallel build constraint and avoid us
hitting the job time limit.

### Details and comments


SooluThomas (approved):
Thank you! 👏

PR #832, opened 2020-02-28
Update LOCALIZATION_CONTRIBUTORS

techwithshadab:
### Summary
Hindi is the national language for India. I would like to translate documentation to Hindi so that it can be easy for Indian folks to understand and work on Qiskit.


nonhermitian:
Can you please move this PR here: https://github.com/Qiskit/qiskit-translations

PR #846, opened 2020-03-15
Add disassemble benchmarks and more permutations to assemble

mtreinish:
### Summary

This commit expands our coverage of the assemble benchmarks by adding
assembly cases with >1 circuit. At the same time this adds a second
benchmark class which adds coverage for qobj disassemble.

### Details and comments

### DIFF
--- a/test/benchmarks/assembler.py
+++ b/test/benchmarks/assembler.py
@@ -34,7 +34,6 @@ class AssemblerBenchmarks:
                                       conditional=True, seed=seed)
         self.circuits = [self.circuit] * number_of_circuits
 
-
     def time_assemble_circuit(self, _, __, ___):
         assemble(self.circuits)
 
--- a/test/benchmarks/assembler.py
+++ b/test/benchmarks/assembler.py
@@ -27,6 +27,7 @@ class AssemblerBenchmarks:
               [1, 5, 10, 50, 100])
     param_names = ['n_qubits', 'depth', 'number of circuits']
     timeout = 600
+    version = 2
 
     def setup(self, n_qubits, depth, number_of_circuits):
         seed = 42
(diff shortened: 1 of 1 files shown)


test/benchmarks/assembler.py
```diff
@@ -16,20 +16,42 @@
...
         seed = 42
         self.circuit = random_circuit(n_qubits, depth, measure=True,
                                       conditional=True, seed=seed)
+        self.circuits = [self.circuit] * number_of_circuits
 
-    def time_assemble_circuit(self, _, __):
-        assemble(self.circuit)
+
+    def time_assemble_circuit(self, _, __, ___):
+        assemble(self.circuits)
```
kdk, line 39: Do we have precedent on whether or not we want to set explicit version numbers or let asv manage it?

test/benchmarks/assembler.py
```diff
@@ -16,20 +16,42 @@
...
-    def setup(self, n_qubits, depth):
+    def setup(self, n_qubits, depth, number_of_circuits):
         seed = 42
         self.circuit = random_circuit(n_qubits, depth, measure=True,
                                       conditional=True, seed=seed)
+        self.circuits = [self.circuit] * number_of_circuits
 
-    def time_assemble_circuit(self, _, __):
-        assemble(self.circuit)
+
```
kdk:
```suggestion
```

kdk (commented):
Looks good, one lint failure, one question.

test/benchmarks/assembler.py
mtreinish: Typically we've been manually versioning things just to be more explicit about it. I'll add a version now

mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/846/commits/fc32f832deb9587349accc85005a2588a812a5e0

kdk (approved):

PR #848, opened 2020-03-16
Add Qiskit Pulse Benchmarks

### LINKED ISSUE #794: Pulse benchmarks
The circuit, transpilation and circuit assembly modules are well benchmarked. There should be associated benchmarks for the pulse path and an associated integration benchmark from the construction of a relatively sophisticated quantum circuit, default backend transpilation and scheduling.

Benchmarks:
- [ ] Schedule construction 
     - with parameters being the:
         - [x] number of unique pulses
         - [x] number of channels
         - [x] ~number of uses of each pulse~
     - options:
        - pulse types:
           - [x] Parametric
           - [x] Sample
       - Building methods:
           - [x] Append instruction
           - [x] Insert instruction right to left
           - [x] Convert the instruction to a schedule and then take the union of all schedules.     
- [ ] Scheduling. Use benchmarks from transpiler but decomposed to a native single qubit gate. Use mockbackend of 20q device:
   - for schedules:
     - [ ] single gate schedule
     - [ ] cx gate schedule
     - [ ] measure schedule
     - [ ] large qasm schedule
  - scheduling methods:
     - [ ] alap
     - [ ] asap
- [ ] End to end benchmark. Add pulse benchmarks to:
     - [ ] quantum volume
     - [ ] randomized benchmarking
- [ ] Assembly
    - [ ] random schedule tests

### LINKED ISSUE #10573: Pulse benchmarks
The circuit, transpilation and circuit assembly modules are well benchmarked. There should be associated benchmarks for the pulse path and an associated integration benchmark from the construction of a relatively sophisticated quantum circuit, default backend transpilation and scheduling.

Benchmarks:
- [ ] Schedule construction 
     - with parameters being the:
         - [x] number of unique pulses
         - [x] number of channels
         - [x] ~number of uses of each pulse~
     - options:
        - pulse types:
           - [x] Parametric
           - [x] Sample
       - Building methods:
           - [x] Append instruction
           - [x] Insert instruction right to left
           - [x] Convert the instruction to a schedule and then take the union of all schedules.     
- [ ] Scheduling. Use benchmarks from transpiler but decomposed to a native single qubit gate. Use mockbackend of 20q device:
   - for schedules:
     - [ ] single gate schedule
     - [ ] cx gate schedule
     - [ ] measure schedule
     - [ ] large qasm schedule
  - scheduling methods:
     - [ ] alap
     - [ ] asap
- [ ] End to end benchmark. Add pulse benchmarks to:
     - [ ] quantum volume
     - [ ] randomized benchmarking
- [ ] Assembly
    - [ ] random schedule tests

SooluThomas:
### Summary

Partially fixes Qiskit/qiskit-terra#10573.

Schedule construction benchmarks:
- with parameters being the:
   - [x] number of unique pulses
   - [x] number of channels
   - [x] ~number of uses of each pulse~
- options:
   - pulse types:
      - [x] Parametric
      - [x] Sample
   - Building methods:
      - [x] Append instruction
      - [x] Insert instruction right to left
      - [x] Convert the instruction to a schedule and then take the union of all schedules.


test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2018, 2019.
```
lcapelluto:
```suggestion
# (C) Copyright IBM 2020.
```

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
...
+    sched = Schedule()
+    for _ in range(number_of_unique_pulses):
+        for channel in range(number_of_channels):
+            sched += my_pulse((DriveChannel(channel)))
+
+    return sched
+
+
+def sample_pulse(number_of_unique_pulses, number_of_channels):
+    my_pulse = SamplePulse(np.random.random(50),
```
lcapelluto: probably `np.random` is fast, but maybe we can hardcode this array so that we avoid including `np.random` in our benchmark of Pulse.

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy as np
+from qiskit.pulse import Schedule, Gaussian, DriveChannel, SamplePulse
+
+
+def build_schedule(my_pulse, number_of_unique_pulses, number_of_channels):
+    sched = Schedule()
+    for _ in range(number_of_unique_pulses):
+        for channel in range(number_of_channels):
+            sched += my_pulse((DriveChannel(channel)))
+
```
lcapelluto: For these short functions, I don't think a new line before the return is necessary. It will help the eye collect 3 short, separate functions if there is less vertical space within each.

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
...
+    return sched
+
+
+def sample_pulse(number_of_unique_pulses, number_of_channels):
+    my_pulse = SamplePulse(np.random.random(50),
+                           name="short_gaussian_pulse")
+
+    return (build_schedule(my_pulse,
```
lcapelluto: superfluous parens here and below

test/benchmarks/pulse/schedule_construction.py
lcapelluto: I see now it looks like it's only called in `setup`. Up to you then!

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
...
+        self.sample_sched = sample_pulse(number_of_unique_pulses,
+                                         number_of_channels)
+        self.parametric_sched = parametric_pulse(number_of_unique_pulses,
+                                                 number_of_channels)
+
+    def time_append_instruction(self, _, __):
+        self.sample_sched.append(self.parametric_sched)
+
+    def time_insert_instruction_left_to_right(self, _, __):
+        if self.parametric_sched.stop_time >= self.sample_sched.start_time:
```
lcapelluto: sample sched start time will always be 0, right?

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,71 @@
...
+def parametric_pulse(number_of_unique_pulses, number_of_channels):
+    my_pulse = Gaussian(duration=25, sigma=4, amp=0.5j)
+
+    return (build_schedule(my_pulse,
+                           number_of_unique_pulses,
+                           number_of_channels))
+
+
+class ScheduleConstructionBench:
+    params = ([1, 2, 5], [8, 128, 2048])
```
lcapelluto, line 43: I might be misunderstanding how these are used. It looks like these will only affect the size of the schedules that are set up, but it won't affect the anticipated runtime of `append` and `insert`. Is that right?

lcapelluto (changes requested):

test/benchmarks/pulse/schedule_construction.py
SooluThomas: In this case, yes.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: afaik, it would affect the runtime if I increase the params values. 

currently, 

```
[ 25.00%] ··· pulse.schedule_construction.ScheduleConstructionBench.time_append_instruction                                                                                                                                                 ok
[ 25.00%] ··· ========================= ========= ========== ==========
              --                               number_of_channels      
              ------------------------- -------------------------------
               number_of_unique_pulses      8        128        2048   
              ========================= ========= ========== ==========
                          1              182±0μs   2.30±0ms   45.2±0ms 
                          2              296±0μs   3.17±0ms   69.2±0ms 
                          5              332±0μs   5.47±0ms   89.0±0ms 
              ========================= ========= ========== ==========

[ 50.00%] ··· pulse.schedule_construction.ScheduleConstructionBench.time_insert_instruction_left_to_right                                                                                                                                   ok
[ 50.00%] ··· ========================= ========= ========== ==========
              --                               number_of_channels      
              ------------------------- -------------------------------
               number_of_unique_pulses      8        128        2048   
              ========================= ========= ========== ==========
                          1              260±0μs   2.78±0ms   47.3±0ms 
                          2              337±0μs   4.46±0ms   87.4±0ms 
                          5              836±0μs   11.1±0ms   214±0ms  
              ========================= ========= ========== ==========
```

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,77 @@
...
+
+def build_schedule(my_pulse, number_of_unique_pulses, number_of_channels):
+    sched = Schedule()
+    for _ in range(number_of_unique_pulses):
+        for channel in range(number_of_channels):
+            sched += my_pulse((DriveChannel(channel)))
+    return sched
+
+
+def sample_pulse(number_of_unique_pulses, number_of_channels):
```
taalexander: Should this be closer to a verb + what it constructs? Something like `build_sample_pulse_schedule`?

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,77 @@
...
+                          number_of_channels)
+
+
+class ScheduleConstructionBench:
+    params = ([1, 2, 5], [8, 128, 2048])
+    param_names = ['number_of_unique_pulses', 'number_of_channels']
+    timeout = 600
+
+    def setup(self, number_of_unique_pulses, number_of_channels):
+        self.empty_sched = sample_pulse(number_of_unique_pulses, 0)
```
taalexander: Won't this amalgamate the timing of constructing all three schedules in each test?

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,77 @@
...
+    def time_append_instruction(self, _, __):
+        self.sample_sched.append(self.parametric_sched)
+
+    def time_insert_instruction_left_to_right(self, _, __):
+        if self.parametric_sched.stop_time >= self.sample_sched.start_time:
+            sched = self.sample_sched.shift(self.parametric_sched.stop_time)
+        sched.insert(self.parametric_sched.start_time,
+                     self.parametric_sched)
+
+    def time_instruction_to_schedule(self, _, __):
```
taalexander: Should this be included in a separate scheduling benchmark?

taalexander (changes requested):

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Right! Since `empty_sched` isn't used in this code, I'll get rid of it. Thanks

test/benchmarks/pulse/schedule_construction.py
SooluThomas: ok. I'll update it

test/benchmarks/pulse/schedule_construction.py
lcapelluto: this is then equivalent to `if True:`?

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,77 @@
...
```
mtreinish: What is the intent of the measurement of this benchmark? Like are you trying to `time inst_map.add()` and `sched.union()`? Right now it's going to include the time of  all the object initialization and access to construct the necessary objects. If you think you want to include all of the time for those pieces then this is fine. But typically we're trying to be a bit more finer grained in what we benchmark.

test/benchmarks/pulse/schedule_construction.py
mtreinish: Yeah, it feels like this will always be a necessary step I would remove the if here.

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,77 @@
...
+
+def build_parametric_pulse_schedule(number_of_unique_pulses,
+                                    number_of_channels):
+    my_pulse = Gaussian(duration=25, sigma=4, amp=0.5j)
+    return build_schedule(my_pulse,
+                          number_of_unique_pulses,
+                          number_of_channels)
+
+
+class ScheduleConstructionBench:
```
mtreinish, line 42: I feel like it would be good to include some benchmarks to build schedules from pulses. Something like: https://github.com/Qiskit/qiskit/blob/master/test/benchmarks/circuit_construction.py#L43-L44

mtreinish (commented):
This is a good start, I've got a couple of questions inline. I also think we should add a benchmarks timing schedule assembly (just calling the build_schedule function). That way we have a baseline on how long it takes to construct pulse schedules as well as all the various modifiers to it.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Okay, I'll remove it.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Yes, it was intended to measure the time taken to do the entire operation. But I now understand that this way of measurement is not really ideal. I'll try to fix it.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Okay, I'll include benchmarks to build sample and parameterized schedules.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Fixed in [4f68e59](https://github.com/Qiskit/qiskit/pull/848/commits/4f68e59810585735e43de503edc00b9a726b3e5f).

test/benchmarks/pulse/schedule_construction.py
mtreinish: Oh, I missed this in my first review, @lcapelluto is correct we should avoid any randomness in the benchmarks, even at setup. You can leverage random functions but only with a fixed seed. If the value here changes between each run it can cause jitter in the measured values over time. something like:

```suggestion
    rng = np.random.RandomState(42)
    pulse_array = rng.random(50)
    my_pulse = SamplePulse(pulse_array,
```

or just do a statically defined array

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,93 @@
...
+                     self.parametric_sched)
+
+    def time_build_instruction(self, _, __):
+        self.inst_map.add('my_pulse', [0], self.parametric_sched)
+
+    def time_instruction_to_schedule(self, _, __):
+        sched = Schedule()
+        sched.union(schedule(self.qc,
+                             self.backend,
+                             inst_map=self.add_inst_map))
```
mtreinish: I'm still a bit confused here; what are we measuring? Is it the time to run `union` or the time to run the scheduler on `self.qc` with the custom cmd def for the opaque gate? The 2 operations don't seem to be linked directly so it's not clear to me why both are being measured by this method.

SooluThomas: It was intended to get the time taken to convert the instruction to a schedule and then take the union of all schedules. (as mentioned in issue - https://github.com/Qiskit/qiskit/issues/794)

mtreinish: Ok sure I see it listed there, but it still doesn't make sense to me. Like are we interested in the performance of `union` or of `schedule` here? The two operations don't seem coupled to me, like we could easily statically define the output of `schedule` and do the union on that instead of invoking the scheduler. That benchmark would be more consistent and make it easier to identify changes in performance over time. Is there something coupling the 2 operations I'm missing? Normally you'd want to measure the time of the 2 operations separately that way when a regression (or an improvement) happens you know what is actually being effected by the change. If they're running together you don't know if it's the scheduler or union causing a regression just by looking at the numbers.

Looking at the other benchmarks here it's `Schedule.union()` we're interested in measuring for this class since everything else here is a synthetic benchmark measuring different constructors of schedules. The performance of these operations should be measured in isolation to ensure consistency. We should save timing for the scheduler to a separate class where we can better characterize it's performance

SooluThomas: Okay. I'll fix it then.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Ok. I'll update it. Thank you!

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def time_insert_instruction_left_to_right(self, _, __):
+        sched = self.sample_sched.shift(self.parametric_sched.stop_time)
+        sched.insert(self.parametric_sched.start_time,
+                     self.parametric_sched)
+
+    def time_build_instruction(self, _, __):
+        self.inst_map.add('my_pulse', [0], self.parametric_sched)
+
+    def time_instruction_to_schedule(self, _, __):
```
mtreinish: Can we do this benchmark in a separate class? The rest of the benchmarks in this class are about bare schedule object construction and modification but this is running the scheduler and has different setup requirements. Normally I'd say it wasn't a big deal, but this has 2 implications in asv, first the grouping in high level view: https://qiskit.github.io/qiskit/ benchmarks are grouped by class. The second is that the setup runtime is impacted by this. `setup()`  gets run for each iteration of the benchmark including each parameter, so the cmd def creation in the setup gets run 54 times for each iteration of the benchmark class (which is typically 4 times, once for cache pre-warming and 3 times to get an error spread). Grouping benchmarks with similar or the same setup in the same class reduces this overhead so we don't balloon in runtime.

mtreinish (commented):
This looks a lot better, thanks for making the updates. Just one more inline comment, we should separate `time_instruction_to_schedule` into another class (maybe in another file). After that I think this is ready.

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Ok. I'll move this to a separate class/file.

test/benchmarks/pulse/schedule_to_instruction_conversion.py
```diff
@@ -0,0 +1,58 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=missing-docstring,invalid-name,no-member
+# pylint: disable=attribute-defined-outside-init
+
+from qiskit import schedule, QuantumCircuit, QuantumRegister
+from qiskit.circuit import Gate
+from qiskit.pulse import Schedule, Gaussian, DriveChannel
+from qiskit.test.mock import FakeOpenPulse2Q
```
mtreinish: I don't think it'll be a huge problem in practice for this case, but it's something to think about if we start using this for more involved benchmarking. The fake backends in terra (even real fake ones like this, not just the ones which are snapshot of real devices) drift over time as we adjust the parameters in the fakes. If we change the pulse defaults in terra this could effect the results of the benchmarks unexpectedly because the fake changed, not because of a change in the `ScheduleToInstructionBench` or the functions being timed. It might be better to have a statically defined cmd def in this module or copy over the fake backend like: https://github.com/Qiskit/qiskit/tree/master/test/benchmarks/backends so we know a change in the backend doesn't change the timing.

test/benchmarks/pulse/schedule_to_instruction_conversion.py
```diff
@@ -0,0 +1,58 @@
...
+        self.add_inst_map = self.inst_map
+        self.add_inst_map.add('my_pulse', [0], self.parametric_sched)
+
+    def time_build_instruction(self, _, __):
+        self.inst_map.add('my_pulse', [0], self.parametric_sched)
+
+    def time_instruction_to_schedule(self, _, __):
+        schedule(self.qc, self.backend, inst_map=self.add_inst_map)
+
+    def time_union_of_schedules(self, _, __):
```
mtreinish: Shouldn't this be in the schedule construction benchmark class instead of the schedule to instruction class?

test/benchmarks/pulse/schedule_construction.py
```diff
@@ -0,0 +1,75 @@
...
+    def time_parametric_pulse_schedule_construction(self,
+                                                    unique_pulses,
+                                                    channels):
+        build_parametric_pulse_schedule(unique_pulses, channels)
+
+    def time_append_instruction(self, _, __):
+        self.sample_sched.append(self.parametric_sched)
+
+    def time_insert_instruction_left_to_right(self, _, __):
+        sched = self.sample_sched.shift(self.parametric_sched.stop_time)
```
mtreinish, line 67: Is the `shift()` call part of what is being measured here? Or is it just the `insert()` with the 2 params?

mtreinish (commented):
A couple more inline comments and one question

test/benchmarks/pulse/schedule_construction.py
SooluThomas: Yes, I think `shift` is part of what is being measured here.

@taalexander Any thoughts?

test/benchmarks/pulse/schedule_to_instruction_conversion.py
SooluThomas: Yes, It should be in the schedule construction benchmark class. Thanks!

SooluThomas: Fixed in [3adcbee](https://github.com/Qiskit/qiskit/pull/848/commits/3adcbee754793109b5c3e1d203af05d6b3e52786)

test/benchmarks/pulse/schedule_to_instruction_conversion.py
SooluThomas: Ok, I'll work on adding the backends.

test/benchmarks/pulse/schedule_construction.py
taalexander: `shift` is also being measured.

test/benchmarks/pulse/schedule_to_instruction_conversion.py
SooluThomas: Fixed in [ff4d41a](https://github.com/Qiskit/qiskit/pull/848/commits/ff4d41a44722e918fca4fdff107d019720ef307d)

mtreinish (commented):
This LGTM, thanks for keeping with this and updating everything based on my review comments. Before we merge it though I ran these new benchmarks locally and the benchmarks emit a ton of deprecation warnings (like so many that it filled up my terminals output buffer and I can't see the results)
```
/home/mtreinish/git/qiskit/qiskit/bench-test/lib/python3.8/site-packages/qiskit/pulse/pulse_lib/pulse.py:47: DeprecationWarning: Calling `Gaussian` with a channel is deprecated. Instantiate the new `Play` instruction directly with a pulse and a channel. In this case, please use: `Play(Gaussian(duration=25, amp=0.5j, sigma=4), DriveChannel(2047))`.
  warnings.warn("Calling `{}` with a channel is deprecated. Instantiate the new `Play` "
```


I understand the the pulse api is in flux and currently being refactored. But what is concerning me here is that the main value of asv is showing performance over time as we make changes. If/when we need to update the benchmark code to adapt to deprecation removals in the future that will invalidate the old results, not only in spirit (by changing the measurement) but the benchmarks are all actually implicitly versioned with the hash of the code. So changing the code treats the benchmark as a new version that can't be compared to the old one. The question I have is while we probably should update the api here to be the non-deprecated version so we can measure consistent results moving forward, is this new api going to stay around for longer than 3 months? Rebuilding the historical data set every time there is an api change won't scale (back-filling old data takes a very long time).

taalexander:
> I understand the the pulse api is in flux and currently being refactored. But what is concerning me here is that the main value of asv is showing performance over time as we make changes. If/when we need to update the benchmark code to adapt to deprecation removals in the future that will invalidate the old results, not only in spirit (by changing the measurement) but the benchmarks are all actually implicitly versioned with the hash of the code. So changing the code treats the benchmark as a new version that can't be compared to the old one. The question I have is while we probably should update the api here to be the non-deprecated version so we can measure consistent results moving forward, is this new api going to stay around for longer than 3 months? Rebuilding the historical data set every time there is an api change won't scale (back-filling old data takes a very long time)

The core content of the tests will remain relevant over time, I imagine this revision of the pulse API should stay current for > 3 months even as we move to the new builder interface which will sit on top of the IR these tests use.

mtreinish:
Ok, sounds good then let's update the API usage in the benchmarks here to avoid the deprecation warnings and then this is good to go from my perspective.

SooluThomas:
Lint error 
```
************* Module test.benchmarks.pulse.schedule_to_instruction_conversion
test/benchmarks/pulse/schedule_to_instruction_conversion.py:20:0: E0611: No name 'Play' in module 'qiskit.pulse' (no-name-in-module)
************* Module test.benchmarks.pulse.schedule_construction
test/benchmarks/pulse/schedule_construction.py:19:0: E0611: No name 'Play' in module 'qiskit.pulse' (no-name-in-module)
```

Will be fixed after Qiskit Terra release.

mtreinish (approved):
LGTM, thanks for keep with it

PR #856, opened 2020-03-17
Integrate Aer benchmarks into Qiskit metapackage

atilag:
### Summary
Integrate Aer benchmarks

Major changes are:
* Created `aer` and `terra` directories and move corresponding benchmarks there.
* Updated `utils.py` to add more common code.
* Homogenize circuit build function names.
* Renamed `terra` benchmark files for consistency.
* Removed `asv.conf.json` and created two different config files for `aer` and `terra` respectively.

### Details and comments
With the goal in mind of having all regression benchmarks in one central point, all
of the AER benchmarks have been moved away from Aer repo to the Qiskit metapackage
repository, where Terra benchmarks reside at the moment. Because there's a lot of
code in common between Terra and Aer benchmarks, I have refactored a bit of code
as part of the integration to gain in clarity and consistency.

Aer benchmarks will run in their own baremetal instance though, to not interfere
with Terra benchmarks, but the results of the benchmark should be published in the
same place, although this is addressed in the PR, it's the script running the benchmarks the one that needs to be updated to do so.

The way to launch each of the benchmarks is now:
```bash
$ asv run --config <asv.aer.conf.json|asv.terra.conf.json> run ...
```

**TODO**
* [x] Test that everything works as expected with Aer benchmarks
* [x] Test that everything works as expected with Terra benchmarks


atilag:
Terra benchmarks are failing with this error:
```bash
virtualenv: error: unrecognized arguments: --no-site-packages
```
Will investigate further tomorrow.

atilag:
Fixed

mtreinish (commented):
So I haven't had a chance to look at everything in depth yet. But from a high level my biggest concern is implicit versioning and the long term history. Asv by default uses a hash of the source for the benchmark as the version: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#benchmark-versioning so by changing the code even if just to update paths or function names we're going to be invalidating the historical data for previous runs. At least for me when I weigh that against unifying the code I think it's more important to keep the historical data than de-duplicating and integrating the benchmark suites. What do you think?

kdk:
> So I haven't had a chance to look at everything in depth yet. But from a high level my biggest concern is implicit versioning and the long term history. Asv by default uses a hash of the source for the benchmark as the version: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#benchmark-versioning so by changing the code even if just to update paths or function names we're going to be invalidating the historical data for previous runs. At least for me when I weigh that against unifying the code I think it's more important to keep the historical data than de-duplicating and integrating the benchmark suites. What do you think?

Can we hard code the existing hash or maybe this can coincide with moving benchmarking hardware?

atilag:
There's a trade-off here between historical data and project organization.

It all depends on how important is the historical data for us.
I'm leaning towards project organization (thus easier maintainability) just because I think the main point of performance regressions benchmarks, is to detect near-term problems in the code being introduced, so only a small portion of the historical data is actually relevant. I don't think comparing with how Qiskit was performing, let's say, 4 releases ago would give us any insights that we haven't knew already.

I have a question though. Does it mean that we need to chose since when we want to restart the benchmarks? ... because running everything since the beginning is non-go situation for me.

transdigiware (approved):

HuangJunye:
@atilag Are you still working on this PR?

atilag:
Oh no. I'm closing this :)

PR #859, opened 2020-03-19
Set up CI with Azure Pipelines

mtreinish:


mtreinish:
Just for an example you can download a zip file of the compiled html docs here for the job from the first run here: https://dev.azure.com/qiskit-ci/qiskit/_build/results?buildId=10059&view=artifacts&type=publishedArtifacts

nonhermitian (approved):

PR #866, opened 2020-03-28
Update Ignis RB Tutorials

ShellyGarion:
### Summary
Update Ignis RB Tutorials for Ignis 0.3 release, due to API changes.


nonhermitian:
A lot of the RB stuff can be put under the module level docstrings in Ignis: https://qiskit.org/documentation/apidoc/verification.html#randomized-benchmarking

I think a singular example on how RB is used for computing gate errors should suffice.

ShellyGarion:
This PR is aimed to update the RB notebook following API changes in ignis, so it is better to merge this in order to prevent the tutorials from failing. 
This is the same PR as in: https://github.com/Qiskit/qiskit-tutorials/pull/885

Later, we can move some of the material and graphs from the tutorials to the docstrings.
This is not trivial, since it requires preparing pre-generated results in order to plot the graphs (since generating and running the RB sequences takes time).

nonhermitian:
This has been reflected in the updated Qiskit tutorials repo.

PR #879, opened 2020-04-10
Remove0.7

jaygambetta:
### Summary
Reading over the install and adding the standard tools that many of the tutorials use.


docs/install.rst
```diff
@@ -86,18 +86,15 @@ Next, install the Qiskit package, which includes Terra, Aer, Ignis, and Aqua.
...
-functions available in Qiskit. You can install these optional
-dependencies by with the following command
+To simplify your experience we also recommend installing the following
 
 .. code:: sh
 
-  pip install qiskit-terra[visualization]
+  pip install matplotlib
+  pip install juypter
+  pip install ipywidgets
```
mtreinish, line 95: This list isn't actually quite enough to run all the visualizations or jupyter widgets. The list of optional dependencies is actually pretty long now: https://github.com/Qiskit/qiskit-terra/blob/master/setup.py#L115-L117

nonhermitian (approved):

PR #881, opened 2020-04-10
Add missing ssh key setup from poBranch update script

mtreinish:
### Summary

In #880 we split the poBranch update script out into a separate job and
script. But as part of that a copy and paste error missed the setup
section for configuring the ssh key needed to actually be able to push
to poBranch. This commit corrects the oversight and adds back the
missing configuration to deploy_translatable_strings.sh so that job can
succeed.

### Details and comments


SooluThomas (approved):
LGTM! Thank you!

Issue #830, opened 2020-02-26
Rename API Docs landing pages


SooluThomas:
The URLs of API References for each element are inconsistent.

### Current 

Terra - https://qiskit.org/documentation/apidoc/qiskit.html
Ignis - https://qiskit.org/documentation/apidoc_legacy/ignis/ignis.html
Aer - https://qiskit.org/documentation/apidoc/qiskit_aer.html
Aqua - https://qiskit.org/documentation/apidoc/index.html
Q Provider - https://qiskit.org/documentation/apidoc/ibmq.html

### Suggested file names and URLs

- [x] Terra - https://qiskit.org/documentation/apidoc/terra.html (apidoc/terra.rst)
- [x] Ignis - https://qiskit.org/documentation/apidoc/ignis.html (apidoc/ignis.rst)
- [x] Aer - https://qiskit.org/documentation/apidoc/aer.html (apidoc/aer.rst)
- [x] Aqua - https://qiskit.org/documentation/apidoc/aqua.html (apidoc/aqua.rst)
- [x] Q Provider - https://qiskit.org/documentation/apidoc/ibmq-provider.html (apidoc/ibmq-provider.rst) - **FIXED: https://github.com/Qiskit/qiskit-ibmq-provider/pull/560**

PR #886, opened 2020-04-16
Bump Meta for Aqua 0.6.6

manoelmarques:
### Summary

Bump the meta repo version to include:

qiskit-aqua==0.6.6

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,7 +48,7 @@ author = 'Qiskit Development Team'
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '0.18.0'
+release = '0.18.1'
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


SooluThomas:
The link to Aqua API references in TOC of [index.rst](https://github.com/Qiskit/qiskit-terra/blob/master/docs/apidocs/terra.rst) should be updated to `Qiskit Aqua <apidoc/aqua>`  as mentioned in #830

mtreinish (changes requested):
Actually, looking at the state of aqua repo at 0.6.6 the change of index in https://github.com/Qiskit/qiskit/pull/886/commits/6d37de6e09eb368a9f85bedd92432c01d0b99985 is incorrect: https://github.com/Qiskit/qiskit-aqua/tree/0.6.6/docs/apidocs (~it's that way on master too~ it's actually correct on master, so we'll have do this for the 0.7 release). We should update it in aqua, but until that's done we can't change the index here. Otherwise it will result in the aqua docs not being in the TOC.

manoelmarques:
I removed the last incorrect commit.

mtreinish (approved):

PR #891, opened 2020-04-21
add accordian to docs

nonhermitian:
### Summary
Also rearrange files so that it matches the terra docs structure.


nonhermitian:
This looks good from the build.

PR #893, opened 2020-04-22
Point to correct circuit lib location

nonhermitian:
### Summary
circuit_library is directly under apidocs


nonhermitian:
Finally done right.

PR #896, opened 2020-04-22
Add benchmarks for bind_parameters

mtreinish:
### Summary

This commit adds a synthetic benchmark for running bind_parameters on a
parameterized circuit. It is basically a dual of the circuit
construction benchmark except that it adds a sweep on the number of
parameters.


kdk (approved):
This looks good. If we have the the runtime, a nice addition would be a benchmark timing `bind_parameters` over a circuit containing composite instructions (to trigger `QuantumCircuit._rebind_definition`), which would be closer to the aqua use-case, and likely quite a bit slower.

mtreinish:
We probably will have the runtime after Qiskit/qiskit-terra#4227 and Qiskit/qiskit-terra#4232 merge. Prior to that though the parameterized circuit construction time will take too long. With 0.13.0 the run time of these benchmarks locally with `asv dev` (so a single iteration) took almost an hour. Actually we might want to either wait or decrease the permutations (or remove the larger parameter numbers) prior to merging this

PR #897, opened 2020-04-23
Meta 0.18.2

jyu00:
### Summary
Bump meta to 0.18.2 for ibmq-provider 0.6.1.


jyu00:
I didn't write a release note because it looks like patch releases don't usually have one? If it's needed I'll make one.

mtreinish (approved):

PR #898, opened 2020-04-23
Add validation error codes

jyu00:


nonhermitian:
#901 makes the docs point to the IQX version of the error codes where they belong.  I have copied the changes here so that they are reflected on the IQX.

PR #912, opened 2020-05-01
Remove typehints plugin to try and fix deprecation warnings

### LINKED ISSUE #911: sphinx-autodoc-typehints causes a deprecation warning on each type hint
### Informations

- **Qiskit version**: All
- **Python version**: All
- **Operating system**: All

Since the sphinx 3.0.0 release the use of sphinx-autodoc-typehints has been leading to runaway deprecation warnings. Every time a type hint is encountered the plugin's use of a deprecated api
emits a deprecation warning like:
```
/home/vsts/work/1/s/.tox/docs/lib/python3.7/site-packages/sphinx/util/inspect.py:555: RemovedInSphinx40Warning: sphinx.util.inspect.Signature() is deprecated
  RemovedInSphinx40Warning)
```

Which took some digging to figure where it was coming from: https://github.com/agronholm/sphinx-autodoc-typehints/blob/a25fa05084b4cee7e9e733603b79e68986a0ffee/sphinx_autodoc_typehints.py#L161 every time that gets called

Sphinx is supposed to have native support for type annotations now s o we should be able 
to remove the plugin usage.

Run a docs build

No deprecation warnings about this

### Suggested solutions

Remove the plugin from the build

mtreinish:
### Summary

This commit removes the autodoc typehints module in an attempt to fix
the deprecation warnings on sphinx >3.0.0 that are raised by every
function signature that uses type hints. Sphinx is supposed to have
native support for typehints this commit is to see if the output from
the docs builds with type hints are formatted decently without the
plugin and to see if the type hints go away.

### Details and comments

Fixes #911

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,7 +63,6 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.mathjax',
     'sphinx.ext.napoleon',
-    'sphinx_autodoc_typehints',
     'sphinx.ext.viewcode',
     'sphinx.ext.extlinks',
     'sphinx_tabs.tabs',


nonhermitian:
This does not seem to work:

from qiskit.org:

Screen Shot 2020-05-01 at 10 57 15

from this PR:

Screen Shot 2020-05-01 at 10 57 26

mtreinish:
hmm, ok I was afraid of that. Then I guess I'll have to contribute to the plugin to update it for sphinx 3.0 to fix the warnings.

PR #918, opened 2020-05-07
Add missing images

nonhermitian:
### Summary
Add images that are missing because I forgot to copy the dir here from terra repo.


nonhermitian:
This works.

PR #938, opened 2020-06-01
Migrate poBranch to Qiskit-translations repository

SooluThomas:
### Summary

This PR adds the changes the script to push the generated translatable files to [qiskit-translations](https://github.com/qiskit-community/qiskit-translations) repository.

### Details and comments

The files that are pushed to the repo are `docs/locale/`, `setup.py`, and `requirements-dev.txt`.

### TO-DO

- [x] Add GitHub keys to push to qiskit-translations repo.

PR #941, opened 2020-06-10
Correct oversight in po sync job

mtreinish:
### Summary

In #938 the process of switching the repo where translations run to
qiskit-community/qiskit-translations was started by updating the deploy
script to push to that repo instead of poBranch on qiskit/qiskit.
However, in that change the script was updated to push the output to a
different repo but missed updating the actual cloning a working copy of
the dest repo. This commit corrects the oversight.


SooluThomas (approved):
Thank you for correcting this.

PR #945, opened 2020-06-16
Add reverse CI job for building tutorials

mtreinish:
### Summary

The qiskit-tutorials repo is using the qiskit metapackage to install
qiskit and running the tutorials in CI. However, we do not have a job in
place to verify that a pending release will not break the tutorials.
Normally we would do this as part of a docs build but because of our
limited time budget we can't do a full doc build and tutorial build in
the same job. This commit just runs the tutorials sphinx build like it
does in the tutorials CI, but uses master from Qiskit/qiskit. This way
we'll catch any issues in a pending release prior to pushing the tag.

### Details and comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,17 @@ jobs:
       env: TOXENV=py38
     - python: "3.5"
       env: TOXENV=lint
+    - python: "3.8"
+      script:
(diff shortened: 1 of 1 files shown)


nonhermitian (approved):
This is exactly what we needed.  Thanks for doing this.

PR #948, opened 2020-06-17
Pivot to github actions instead of travis for docs publish

mtreinish:
### Summary

The documentation builds and upload time for the publish job can
consistently take well over 1 hour. The travis jobs have a hard limit
of 50min and this is similar in other CI systems like azure-pipelines
(60 mins). According to the docs for github actions the job
timeout is 6 hours [1] which should be more than enough for the docs
builds. This commit adds an initial workflow for publish docs and
removes the travis job config for publishing the docs so we will
rely on a post merge action to build and publish the documentation
moving forward.

### Details and comments

[1] https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,27 @@
+name: Docs Publish
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


SooluThomas (approved):
LGTM! Thank you!

PR #954, opened 2020-06-22
fix double widgets

nonhermitian:
### Summary
The widget javascript is being loaded twice.  This takes the solution from here: https://github.com/spatialaudio/nbsphinx/issues/378


nonhermitian:
Yay, this worked.  Price to pay is no widgets in tutorials, but that fine.

mtreinish (approved):

PR #957, opened 2020-06-28
Updated docs/install.rst. Fixes issue #894

### LINKED ISSUE #894: Conda environment management command

- What is the file name(s) where you found the bug?
https://github.com/Qiskit/qiskit/blob/master/docs/install.rst

- What is the current text that needs to be fixed or removed? (You can copy and paste)
The current text uses `source activate name_of_my_env` for linux and mac and `activate name_of_my_env` for windows to activate a conda environment.

- What should the documentation say instead?
Since conda 4.6, it seems to use `conda activate` to activate an environment.
https://docs.conda.io/projects/conda/en/latest/commands.html#conda-vs-pip-vs-virtualenv-commands
I'm afraid that I didn't test it because I'm a virtualenv user. 

- Are there other related issues or pull requests about this bug?
I have no idea.

- Which files need to be updated?
https://github.com/Qiskit/qiskit/blob/master/docs/install.rst

- What content needs to be updated or replaced? (You can copy and paste)
- What new content needs to be added?
If qiskit supports only conda 4.6 or later, it seems good to replace `source activate name_of_my_env` and `activate name_of_my_env` with `conda activate name_of_my_env`; but if qiskit supports 4.5 or before, it is necessary to keep the current text.

Ritvik-Sapra:
### Summary
The file "install.rst" in docs explained creating and activating a anaconda environment. It showed the following command.
`source activate name_of_my_env`
However, for new anaconda version (>4.6), this command is depreciated and instead following command is used.
`conda activate name_of_my_env`

This has been updated in the file.

### Details and comments
Working on qiskit using anaconda is very important. More than virtual environments, anaconda is preferred by many people. `source activate` or `activate` is a command for older versions of anaconda. New versions are depreciating this command and for following versions, this command might not even be recognized. Hence, _this pull request updates the docs_ by giving instructions for new version.
While installing, anaconda still gets installed with older versions and many people may not update it. Therefore, the _older command is not deleted_. Just new command is added for current and future version of anacoda.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -160,7 +160,7 @@ returns only the version for the ``qiskit-terra`` package. This is because
 the ``qiskit`` namespace in Python doesn't come from the Qiskit package, but
 instead is part of the ``qiskit-terra`` package.
 
-.. jupyter-execute::
+.. code::
 
    import qiskit
    qiskit.__version__
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -33,37 +33,28 @@ included with Anaconda. A Conda environment allows you to specify a specific
 version of Python and set of libraries. Open a terminal window in the directory
 where you want to work.
 
+It is preferred is you use Conda prompt installed with the Anaconda. 
+All you have to do is create a virtual environment in Anaconda and activate the environment. 
+These commands can be run in Anaconda prompt irrespective of Windows or linux machine.
+
 Create a minimal environment with only Python installed in it.
 
 .. code:: sh
 
   conda create -n name_of_my_env python=3
 
-If you are using conda version 4.6 or above, which is recommended, use following command to activate your environment:
+Activate your new environment.
+.. code:: sh
 
   conda activate name_of_my_env
-This command can be used irrespective of Windows or Unix/Linux. Simply type these commands in anaconda prompt.
 
 
-.. code:: sh
+.. note::
+  If you are using conda versions prior to 4.6, use:
+   
+    For Linux or macOS: ``source activate name_of_my_env``
+    For Windows: ``activate name_of_my_env``
 
-  source activate name_of_my_env
-
-Or, if you're using Windows
-
-1. Install Anaconda
-2. Search for Anaconda Prompt
-3. Open Anaconda Prompt
-
-Use the following commands
-
-.. code:: sh
-
-  conda create -n name_of_my_env python=3
-
-.. code:: sh
-
-  activate name_of_my_env
 
 
 Next, install the Qiskit package, which includes Terra, Aer, Ignis, and Aqua.
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,12 +44,14 @@ Create a minimal environment with only Python installed in it.
   conda create -n name_of_my_env python=3
 
 Activate your new environment.
+
 .. code:: sh
 
   conda activate name_of_my_env
 
 
 .. note::
+
   If you are using conda versions prior to 4.6, use:
    
     For Linux or macOS: ``source activate name_of_my_env``
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -160,7 +160,7 @@ returns only the version for the ``qiskit-terra`` package. This is because
 the ``qiskit`` namespace in Python doesn't come from the Qiskit package, but
 instead is part of the ``qiskit-terra`` package.
 
-.. jupyter-executive::
+.. jupyter-execute::
 
    import qiskit
    qiskit.__version__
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -39,6 +39,11 @@ Create a minimal environment with only Python installed in it.
 
   conda create -n name_of_my_env python=3
 
+If you are using conda version 4.6 or above, which is recommended, use following command to activate your environment:
+
+  conda activate name_of_my_env
+This command can be used irrespective of Windows or Unix/Linux. Simply type these commands in anaconda prompt.
+
```
SooluThomas: I think these lines should be added after 
```rst
.. code:: sh

   activate name_of_me_env
```
or put the older commands (used for conda versions prior to 4.6) in a note. 
```rst
.. note::
   
   If you are using conda versions prior to 4.6, use:
   
   For Linux or macOS: ``source activate name_of_my_env``
   For Windows: ``activate name_of_my_env``
```

And add Windows instructions after/in L33-L34. Maybe something like:
```rst
If you are using Linux or macOS, open a terminal window in the directory where you want to work. 
For windows, open Anaconda Prompt.
```

Also, to fix the lint error, each line should not be more than 100 characters.

SooluThomas (changes requested):
Thanks for updating this. I have a few suggestions:

docs/install.rst
Ritvik-Sapra: Yes, I actually did this before. But then updated the commit and put it before because mostly new versions will be used. So any user will feel easier to find these instructions. However, your suggested way makes it even more clean. 
I updated the required changes.

Ritvik-Sapra:
image
I am getting this error when i'm trying to run ``tox -elint``. 

I first installed tox using ``pip install tox`` with cmd. and I installed pycodestyle. But its still showing this error.

SooluThomas:
I don't think the change made in the commit https://github.com/Qiskit/qiskit/pull/957/commits/bf329ee19c9a5b16128ce681b35f6a237dd65c00 is necessary. Please revert it and you can also check here - https://travis-ci.com/github/Qiskit/qiskit/jobs/355186912#L470 to see the lines which are failing lint tests.

nonhermitian (commented):
Why are we removing the jupyter execute here?

Ritvik-Sapra:
Hey, yes. I will revert back to jupyter executive. @nonhermitian  Actually that line kept showing as an error in VS code. So I tried installing Sphinx and then there were more errors. I will revert back all the changes and then run the code again in a conda env now as @SooluThomas guided..

Ritvik-Sapra:
image
I reverted back jupyter-executive and removed whitespaces from 36 and 37 as told in [this link](https://travis-ci.com/github/Qiskit/qiskit/jobs/355186912#L470) @SooluThomas 
But this error shown above is being given by VS code. This is why I removed it in the first place. Any idea why? 😐

docs/install.rst
```diff
@@ -96,7 +94,7 @@ versions of macOS) you'll need to put ``qiskit[visualization]`` in quotes:
   After you've installed and verified the Qiskit packages you want to use, import
   them into your environment with Python to begin working.
 
-.. code:: python
+.. jupyter-executive:: python
```
SooluThomas:
```suggestion
.. jupyter-execute::
```

docs/install.rst
```diff
@@ -162,7 +160,7 @@ returns only the version for the ``qiskit-terra`` package. This is because
 the ``qiskit`` namespace in Python doesn't come from the Qiskit package, but
 instead is part of the ``qiskit-terra`` package.
 
-.. jupyter-execute::
+.. jupyter-executive::
```
SooluThomas:
```suggestion
.. jupyter-execute::
```

SooluThomas (changes requested):
It is not ``jupyter-executive``. It's jupyter-execute.

Ritvik-Sapra:
Done. But that VS code is still showing error in this line.

SooluThomas:
I'm not sure why your VSCode is showing the error. But I think this PR is good to be merged.

docs/install.rst
```diff
@@ -96,7 +94,7 @@ versions of macOS) you'll need to put ``qiskit[visualization]`` in quotes:
   After you've installed and verified the Qiskit packages you want to use, import
   them into your environment with Python to begin working.
 
-.. code:: python
+.. jupyter-execute::
```
SooluThomas:
```suggestion
.. code:: python
```

SooluThomas (approved):
LGTM! Thanks for working on this!

PR #960, opened 2020-07-10
Prepare for publishing Korean Qiskit Documentation

SooluThomas:
### Summary

This PR adds ko_KR to the configuration to prepare for publishing Korean Documentation

### DIFF
--- a/docs/versionutils.py
+++ b/docs/versionutils.py
@@ -29,6 +29,7 @@ translations_list = [
     ('en', 'English'),
     ('ja_JP', 'Japanese'),
     ('de_DE', 'German'),
+    ('ko_KR', 'Korean')
 ]
 
 default_language = 'en'


mtreinish:
Were you think that we should merge qiskit-community/qiskit-translations#118 before this so that when we add the link to the docs the korean content is up? Either way I think there is a race condition because if we merge this first then people could navigate to the korean translation and get a 404. But if we merge qiskit-community/qiskit-translations#118 first then the translated docs (including korean) won't know about the korean translations and you won't be able to navigate to it until after a rebuild of the translated docs (once this merges).

SooluThomas:
I think either ways I'm expecting a small lag between adding the link to Korean docs and actually getting the Korean docs. So, it might be better to merge https://github.com/qiskit-community/qiskit-translations/pull/111 first so that we won't get a 404.

mtreinish (approved):
Ok, sounds like a plan qiskit-community/qiskit-translations#118 has merged. Once the build finishes and is uploaded to COS lets merge this and go from there

PR #961, opened 2020-07-10
Trigger docs build for updated optimization tutorials

nonhermitian:
### Summary
The optimization tutorials were updated to include new gallery images and the notebooks re executed again so as the version shows the latest Qiskit version (given we have more time, in the future we should execute the notebooks in this docs build so as the content shows up with lastest version info).


nonhermitian:
This worked.  It is less than ideal, but can worry about that later.

PR #973, opened 2020-07-17
Cap matplotlib and add set -e to deploy script

mtreinish:
### Summary

The recent matplotlib broke the hinton plotter (which has been fixed on
terra master) and the documentation fails to build now when it runs the
docs for the hinton plot function. Because the deploy script was missing
a set -e the failed documentation build didn't stop the script and it
continued to run a sync on empty directory which lead to the
documentation getting deleted. This commit fixes the docs builds by
pinning matplotlib until the next terra release and setting set -e on
the deploy script so we the documentation doesn't accidentally get deleted
when the docs build fails.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,7 +7,7 @@ sphinx-tabs>=1.1.11
 sphinx-automodapi
 jupyter
 jupyter-sphinx
-matplotlib>=2.1
+matplotlib>=2.1,<3.3.0
 pydot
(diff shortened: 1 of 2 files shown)


SooluThomas (approved):
Thank you

mtreinish:
This fixes the docs builds and we need to get this out to restore the documentation so I'm going to merge as is. The tutorials job is still broken because matplotlib doesn't get capped there so I'll have to fix that after this merges

PR #977, opened 2020-07-22
Pre-requisite in installing Qiskit in an empty conda env

### LINKED ISSUE #696: Cannot pip install qiskit in empty conda env
### Information

- **Qiskit version**: 0.13.0
- **Python version**: 3.7 
- **Operating system**: OSX

```
ModuleNotFoundError: No module named 'numpy'
    **************************************************
    * numpy was not installed in your system.  Please run
    *     pip install numpy
    * before installing pyscf.
    **************************************************
```
However this is exactly what we tell people to do in the install instructions: https://qiskit.org/documentation/install.html

Ritvik-Sapra:
### Summary
As mentioned in Issue #696 when installing Qiskit in a clean environment, it popped up error saying "Numpy was not installed, please install it first". However, no instructions about this is given on the website or in the [install](https://qiskit.org/documentation/install.html) section. Therefore, this PR changes the file [install.rst](https://github.com/Qiskit/qiskit/blob/master/docs/install.rst) accordingly. 

- A section of Pre-requisite was added, to make the user aware about at least some python packages that must be there.

- Numpy and Matplotlib was added as minimum pre-requisite

### Details and comments
Qiskit makes use of multiple python packages internally. When installing in a clean conda env, it will show error since internally, certain packages like, numpy will fail to import. Some information regarding pre-requisite must be given to the user.


Ritvik-Sapra:
The test of building docs to HTML is failing. Is there any way to know where the error is? 
@nonhermitian @SooluThomas

nonhermitian:
Here is the error:

```
Notebook error:
NotebookValidationError in tutorials/noise/3_measurement_error_mitigation.ipynb:
'execution_count' is a required property

Failed validating 'required' in execute_result:

On instance['cells'][66]['outputs'][0]:
{'data': {'text/html': '<h3>Version Information</h3><table><tr><th>Qiskit '
                       'Software</th><...',
          'text/plain': '<IPython.core.display.HTML object>'},
 'metadata': {},
 'output_type': 'execute_result'}

```

Looks like some issue in the tutorials.

Ritvik-Sapra:
I think this is a separate issue. Should I open another PR ?

Ritvik-Sapra:
@nonhermitian @SooluThomas all checks finally passed. Could you please review if my changes are adequate?

Ritvik-Sapra:
Are the checks okay now ? @nonhermitian

SooluThomas:
@Ritvik-Sapra This has been fixed in the PR https://github.com/Qiskit/qiskit-tutorials/pull/970

Your ci checks are successful now.

Ritvik-Sapra:
@SooluThomas oh ok cool. Should there be any changes in this PR? @nonhermitian

nonhermitian (commented):
So my opinion is that matplotlib should be dropped here since it is optional. Also I wouldn’t say “these are the important ones”. I would just saw the follow package needs to be preinstalled. We should probably also give the pip install command with the versioning that matches setup.py. @mtreinish your thoughts?  For later releases perhaps we do the same trick we do now for cython for the missing NumPy case. If so then this section can be removed in the future.

mtreinish (changes requested):
This is not an accurate addition to the documentation, Neither matplotlib or numpy are prerequisites of qiskit, numpy is a runtime dependency and matplotlib isn't even required. numpy will be installed by pip as a required dependency when `pip install qiskit` is run. You only need matplotlib to use visualizations which is covered below in the documentation. A prerequisites here would be something like having a computer with an internet connection.

The issue in #696 is because pyscf is poorly packaged and doesn't include binaries on windows. The error there is caused by pyscf trying to compile and not properly having numpy listed as a build requirement. That being said there isn't much we can do about this except removing pyscf from the aqua requirements, which probably isn't a bad idea. It's already optional because of the lack of windows binaries. I'm not sure it's something we can or should address with documentation

nonhermitian:
I am confused. We need to special case cython, and the cython code requires numpy to build in our case.

mtreinish:
> I am confused. We need to special case cython, and the cython code requires numpy to build in our case.

If you're referring to this section:

https://github.com/Qiskit/qiskit-terra/blob/master/setup.py#L20-L25

that is needed because we call cython in the setup.py and we can't express cython as a build dependency if python fails with an import error loading the setup.py trying to import cython before we can express that dependency to setuptools or pip. The numpy dependency for qiskit doesn't have that issue because we don't use it in the setup.py. You can easily test this today by just running `pip install --no-binary qiskit-terra qiskit-terra` in a fresh env which will build the package from source instead of using the wheel. For wheels it doesn't matter because we've precompiled the binaries (unless you're on non-x86 where there are no terra wheels).

The error message reported in #696 isn't actually coming from qiskit it's from pyscf (it even says that, that's how I knew it was the problem: `before installing pyscf.`)

nonhermitian:
Ok Yes I get it now. Then I guess we can close this in favor of an optional pyscf.

Ritvik-Sapra:
Ok, thank you for everyone's input. I will close this then. Maybe we should close the corresponding issue too. Thanks again !

PR #989, opened 2020-08-03
Make executing notebooks env configurable and set for deploy job

mtreinish:
### Summary

This commit adds support for configuring the execution of the
tutorials via an env variable 'QISKIT_DOCS_BUILD_TUTORIALS'. When set
the value is used as the nbsphinx_execute value in the sphinx config,
this will let users or CI jobs change the default value of 'never' to
enable executing the tutorials as part of the documentation builds. The
default of 'never' was used because we do not have a sufficient CI
time budget in test jobs for running a full doc build and executing the
tutorials. However, the publish job running in gha has a 6 hour time
limit and doesn't have the slow feedback time concern so we can let it
spin building documentation. Having the sphinx build run the notebook
provides a uniform environment for all the notebook output and it
ensures that environment differences from tutorial contributors don't
lead to inconsistent rendered documentation.

### Details and comments

We already have a tutorials job that executes the notebooks on PRs to
ensure that they always work. So the docs publish jobs shouldn't be
wedged by a broken notebook.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,5 +23,6 @@ jobs:
       env:
         encrypted_rclone_key: ${{ secrets.encrypted_rclone_key }}
         encrypted_rclone_iv: ${{ secrets.encrypted_rclone_iv }}
+        QISKIT_DOCS_BUILD_TUTORIALS: 'always'
       run: |
         tools/deploy_documentation.sh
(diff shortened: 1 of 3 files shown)


nonhermitian:
This is great. This was the last missing piece in the workflow. Now the notebooks will all show same version for outputs.

nonhermitian (approved):

Issue #993, opened 2020-08-05
Images not found in install.rst


prasad-kumkar:
Browser:
Google Chrome, Safari

- Which files need to be updated?
https://github.com/Qiskit/qiskit/blob/master/docs/install.rst

- What content needs to be updated or replaced? 
The image needs to be updated under point 2 and 3 of Access IBM quantum systems:
https://github.com/Qiskit/qiskit/blob/master/docs/install.rst#access-ibm-quantum-systems
<img width="1198" alt="Screenshot 2020-08-05 at 10 37 20 AM" src="https://user-images.githubusercontent.com/43164022/89374764-eb43fb80-d709-11ea-8916-402d5a4d78ca.png">

- What new content needs to be added?
Image of where to find the section 'My accounts'
Image of where to get an API token.

mtreinish:
These images are correct, the documentation is not designed for viewing via github's builtin rst viewer but instead for use with sphinx to build the documentation. We host the html compiled version of the documentation here:
https://qiskit.org/documentation/install.html#access-ibm-quantum-systems
which you can see has the images correctly.

prasad-kumkar:
Oh okay. Thank you for sharing that @mtreinish, I didn't know about that. I will close this issue now.

PR #994, opened 2020-08-05
Update install.rst

gunchamalik:
fix broken image links mentioned in #993 

✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.

### Summary
fix broken image links mentioned in #993 

### Details and comments
The required images already exist in the specified folder. The .rst file had the path to images starting with "/". I have removed the first character "/" of the path to pick the images from the right path.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -126,14 +126,14 @@ To configure your account, you create a local configuration file which includes 
 
 **2.**  Navigate to **My Account** to view your account settings.
 
-.. image:: /images/figures/install_0.png
+.. image:: images/figures/install_0.png
    :alt: Image of where to find the section 'My accounts'.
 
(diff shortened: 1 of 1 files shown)


mtreinish:
These images are correct, the documentation is not designed for viewing via github's builtin rst viewer but instead for use with sphinx to build a rendered version of documentation. We host the html compiled version of the documentation here:
https://qiskit.org/documentation/install.html#access-ibm-quantum-systems
which you can see has the images correctly. The paths in the rst are for sphinx to use the images during builds

gunchamalik:
ok.. thanks for sharing these details @mtreinish .In that case, should this pull request and the corresponding defect be closed?

SooluThomas:
Yes, I'll close this issue.

PR #999, opened 2020-08-06
Add release notes for provider 0.8

jyu00:
### Summary
Add release notes for provider 0.8.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1292,11 +1292,6 @@ Aqua 0.7.4
 IBM Q Provider 0.8.0
 ====================
 
-.. _Release Notes_0.8.0:
-
-0.8.0
-=====
-
 .. _Release Notes_0.8.0_New Features:
 
 New Features
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -1292,6 +1292,141 @@ Aqua 0.7.4
 IBM Q Provider 0.8.0
 ====================
 
+.. _Release Notes_0.8.0:
+
+0.8.0
+=====
+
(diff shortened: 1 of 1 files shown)


docs/release_notes.rst
```diff
@@ -1292,6 +1292,136 @@ Aqua 0.7.4
 IBM Q Provider 0.8.0
 ====================
 
+.. _Release Notes_0.8.0_New Features:
+
+New Features
+------------
+
+.. releasenotes/notes/0.8/backend-reservation-c445db6d8de5a896.yaml @ b'ced9a2f60f9d27e907e0d1e47e0a2f28d531555e'
```
mtreinish, line 1300: In the past these lines caused a sphinx failure. I normally remove them manually. You'll probably need to remove them too, we can wait for CI to get back (which might be a while because of the backlog). Or you can do a sphinx build locally to test

mtreinish (commented):
This lgtm, but I think the `.. releasenotes` lines will cause a failure

docs/release_notes.rst
jyu00: I was able to build it locally... so I'll wait and see if CI fails before making the change.

mtreinish (approved):
Ok cool, yeah the docs jobs are passing. I wish I had known that they worked now (probably when we first started using reno it was with an old sphinx version or the reno plugin wasn't installed) and didn't waste my time deleting all those lines :)

mtreinish:
I just hit the failure on the CI job for bump_meta, it's line length for docs lint. The lines are >100 characters line. It's probably a bug in doc8 since it should ignore directives and comments. I'll open a bug with doc8

PR #1014, opened 2020-08-12
Update requirements-dev.txt

d4rkc0nd0r:
### Summary
I have changed pydot to pydotplus.

### Details and comments
Pydotplus is an improved version of pydot, that provides python interface to Graphviz's Dot language.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,7 +8,7 @@ sphinx-automodapi
 jupyter
 jupyter-sphinx
 matplotlib>=2.1,<3.3.0
-pydot
+pydotplus
 ipywidgets>=7.3.
 pillow>=4.2.1
 seaborn>=0.9.0


mtreinish (changes requested):
Thanks for opening a PR, unfortunately the `requirements-dev.txt` isn't used in this manner. It's used to provide a list of dependencies needed for running development tools and tests in this repo that are outside the set of general project requirements. Since this repo just contains a metapackage configuration, documentation and some minimal integration testing this is basically there to specify dependencies for building documentation. In the case of pydot this is an optional requirement for visualization in qiskit-terra and is used by several graph drawers. These drawers get invoked by the documentation build and require that pydot is installed to function. That is why the docs job failed here. If you want to update the graph drawers I would suggest opening an issue or PR against Qiskit/qiskit-terra. For now I'm going to close this since it won't work, feel free to reopen if you think I'm missing something.

PR #1015, opened 2020-08-12
Update requirements-dev.txt

d4rkc0nd0r:
### Summary
I have added Scipy.

### Details and comments
Scipy contains modules for various commonly used tasks in scientific programming, including linear algebra, integration (calculus), ordinary differential equation solving, and signal processing, which is really helpful in Quantum Programming.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -15,3 +15,4 @@ seaborn>=0.9.0
 nbsphinx
 cvxpy<1.1.0
 pyscf<1.7.4
+scipy


mtreinish (changes requested):
Thanks for opening a PR, unfortunately the `requirements-dev.txt` isn't used in this manner. It's used to provide a list of dependencies needed for running development tools and tests in this repo that are outside the set of general project requirements. Since this repo just contains a metapackage configuration, documentation and some minimal integration testing this is basically there to specify dependencies for building documentation. In this case scipy is a requirement for installing qiskit and will already be installed so it doesn't need to be in the requirements-dev.txt list. Since this shouldn't be added to the list I'm going to close this. Feel free to reopen if you think I'm missing something

PR #1019, opened 2020-08-13
Remove redundant docs job from travis

SooluThomas:
### Summary

Both the travis CI tests and azure builds docs in every opened PR. The travis docs job runs for approx 20 minutes and azure for 47 minutes. This PR intends to remove one of the jobs to save some CI tests time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,17 +29,6 @@ jobs:
       env: TOXENV=py38
     - python: "3.5"
       env: TOXENV=lint
-    - python: "3.8"
-      script:
-        - git clone https://github.com/Qiskit/qiskit-tutorials
-        - python -m pip install --upgrade pip setuptools wheel virtualenv
-        - virtualenv /tmp/docs_build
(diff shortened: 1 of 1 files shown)


mtreinish (changes requested):
This job is actually needed. The docs job we run in azure pipelines does not actually execute the tutorials it just relies on the per-rendered output cells in the notebook files. This is done for execution time because as you point out in the PR summary we're approaching the 1 hour job time limit in azure already. We do actually execute the notebooks in the publish job though see #989 for the details on that. This ci job clones the tutorials repo and just executes the tutorials. This is necessary to ensure that any changes we make in the metapackage do not break any tutorials. It serves as 2 things, an upgrade test to ensure that potential new metapackage releases do not regress and also an integration test for new version combinations. The use of sphinx is more just an artifact of how the tutorials are packaged think of it more of a unittest job than a docs test. I'm going to close this since we need this job feel free to reopen if I'm missing something

PR #1028, opened 2020-08-19
Fix grammar in install.rst

RafeyIqbalRahman:
### Summary
Fixed 3 grammatical errors in install.rst

### Details and comments
 The fixes are as follows:

- Changed "If the packages installed correctly ..." to "If the packages **are** installed correctly ...".
- Changed the phrase "It is worth point at ..." to "It is worth pointing that ...". Now it makes more sense.
- Added a comma to respect punctuation in a long, complex sentence.


SooluThomas (dismissed):
LGTM! Thank you

3yakuya:
Very minor note, but I actually think "If packages installed (...)" is somehow more accurate, as it reflects the fact that they were supposed to be installed _just before _ user reads this bit.

Alternative change would be "got installed", though referring to software as "installing itself" is also somehow accepted (we often say "system is updating" or "packages are installing".)

RafeyIqbalRahman:
> Very minor note, but I actually think "If packages installed (...)" is somehow more accurate, as it reflects the fact that they were supposed to be installed _just before _ user reads this bit.
> 
> Alternative change would be "got installed", though referring to software as "installing itself" is also somehow accepted (we often say "system is updating" or "packages are installing".)

I went with "got installed".

docs/install.rst
```diff
@@ -82,8 +82,8 @@ recommended to install qiskit with the visualization extra requirements
 
   pip install qiskit[visualization]
 
-It is worth point at if you're a zsh user (which is the default shell on newer
-versions of macOS) you'll need to put ``qiskit[visualization]`` in quotes:
+It is worth pointing that if you're a zsh user (which is the default shell on newer
```
mtreinish:
```suggestion
It is worth pointing out that if you're a zsh user (which is the default shell on newer
```

docs/install.rst
```diff
@@ -72,7 +72,7 @@ Next, install the Qiskit package, which includes Terra, Aer, Ignis, and Aqua.
   command will attempt to install qiskit-aer from sdist (source distribution)
   which will try to compile aer locally under the covers.
 
-If the packages installed correctly, you can run ``conda list`` to see the active
+If the packages got installed correctly, you can run ``conda list`` to see the active
```
mtreinish:
```suggestion
If the packages were installed correctly, you can run ``conda list`` to see the active
```

3yakuya (approved):

mtreinish (approved):

PR #1038, opened 2020-09-08
Migrate CI to github actions

mtreinish:
### Summary

Since the period with an extra 10 concurrent job on travis is ending and
there is already extensive resource contention in the shared azure
pipelines instance for the qiskit organization we need to find an
alternative for CI to maintain job throughput. This commit migrates all
the existing CI jobs to github actions which has a very generous quota
for open source projects.

PR #1051, opened 2020-10-02
Add back translatable string job

mtreinish:
### Summary

In #1038 we migrated the Qiskit metapackage CI to use Github Actions
instead of Travis CI. However, as part of that migration one job was
overlooked and not migrated from Travis to Github Actions, the
translatable string post-merge job which generates the po files and
updates the qiskit-community/qiskit-translations repo which new
translatable strings in the docs. The translation repo hasn't
received string updates since we migrated to github actions because
in #1038 we accidently deleted the job and didn't add an equivalent
to Github Actions. This commit corrects the oversight and adds back
a translatable string job to Github Actions.

### Details and comments

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,3 +26,25 @@ jobs:
         QISKIT_DOCS_BUILD_TUTORIALS: 'always'
       run: |
         tools/deploy_documentation.sh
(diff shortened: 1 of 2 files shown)


SooluThomas (approved):
Thanks for catching this

PR #1066, opened 2020-10-19
Update Chemistry 0.8 renos

Cryoris:
### Summary

Fix the chemistry renos for the 0.8 release where the description was either missing or placed in the wrong paragraph in the reno file.


mtreinish (approved):
This looks fine, but in the future all the class names used here are better as internal references to  the documentation to make navigation to the full docs for a class easier. For example replacing:

```
``NumpyEigensolver``
```

with

```
:class:`~qiskit.aqua.algorithms.NumPyEigensolver`
```

The formatting will look nearly identical but it will creat a link to the documentation for the class.

PR #1072, opened 2020-10-22
Qualitative transpiler benchmarks using different layouts, routing methods, optimization levels

adekusar-drl:
### Summary
Added a benchmark suite that measures transpilation time and output depth on some circuits that may be transpiled better and faster. A choice of circuits is based on a more thorough comparison.

### Details and comments
- Added 3 new circuits to measure transpilation time
- Added 3 new other circuits to measure output depth of the circuits.
- The benchmark suite runs the transpiler using optimization levels `[2, 3]`, routing methods: `["stochastic", "sabre"]`, and layout methods: `["dense", "noise_adaptive", "sabre"]`.


test/benchmarks/transpiler_qualitative.py
```diff
@@ -0,0 +1,98 @@
...
+
+import os
+
+from qiskit import QuantumCircuit
+from qiskit.compiler import transpile
+from qiskit.test.mock import FakeToronto
+
+
+class TranspilerQualitativeBench:
+    params = ([2, 3],
```
kdk: Is there a reason not to include levels 0 and 1 here? The performance would be worse (the runtime should probably be better), but it might still be interesting to track changes over time.

test/benchmarks/transpiler_qualitative.py
```diff
@@ -0,0 +1,98 @@
...
+            os.path.join(self.qasm_path, "depth_mod8-10_178.qasm"))
+
+        self.time_qc1 = QuantumCircuit.from_qasm_file(
+            os.path.join(self.qasm_path, "time_cnt3-5_179.qasm"))
+        self.time_qc2 = QuantumCircuit.from_qasm_file(
+            os.path.join(self.qasm_path, "time_cnt3-5_180.qasm"))
+        self.time_qc3 = QuantumCircuit.from_qasm_file(
+            os.path.join(self.qasm_path, "time_qft_16.qasm"))
+
+    def track_depth_transpile_qc1(self, optimization_level, routing_method,
```
kdk: It might be good to include the name of the circuit (e.g. depth_4gt10-v1_81) instead of qc_1 in the function name, since the function name is what will show up in the benchmark UI.

kdk (commented):
Looks good. Do you know the total current runtime of the new benchmarks? @mtreinish Do you know how much headroom we have in terms of runtime on our current benchmarks per commit?

mtreinish:
> Looks good. Do you know the total current runtime of the new benchmarks? @mtreinish Do you know how much headroom we have in terms of runtime on our current benchmarks per commit?

Things changed a bit recently since we merged #1069 and we don't have a great data set yet because things have been pretty quiet the past few days.

Looking at the runs since we merged #1069:

| Date | Number of Commits to Terra | Push timestamp for results |
|--------|---------------------------------------|--------------------------------------|
| 10/22 | 3 | 06:25 |
| 10/23 | 1 | 03:08 | 
| 10/24 | 0 | 00:01 |
| 10/25 | 0 | 00:01 |
| 10/26 | 0 | 00:01 |
| 10/27 | 4 | 09:19 |

so it looks like our runtime is at ~2.15 to 3 hours per commit since we dropped the peakmem benchmarks. Which is definitely down from where it was before. It looks like #1069 saved us about an hour of runtime (for example the 10/20 run published results at 13:07 for 4 commits).

mtreinish:
I ran the 6 benchmarks here locally this morning it took ~15min (which included the environment build time). So we definitely have room for these (and probably also if we add levels 0 and 1 too).

test/benchmarks/transpiler_qualitative.py
adekusar-drl: No reason except the run time. Added 0 and 1. If it's too slow I can keep just a couple of levels.

test/benchmarks/transpiler_qualitative.py
adekusar-drl: I was thinking of that, but decided that the names would be too ugly. So agree, now I have fancy names.

adekusar-drl:
@kdk With levels 0 and 1:
- `asv dev --bench TranspilerQualitativeBench` took 0:15:21.03 (15 minutes) on my laptop
- `asv run --bench TranspilerQualitativeBench` took 0:41:8.94

adekusar-drl:
Can anybody restart the CI? There's a timeout in the tutorials.

kdk (approved):
Looks good, thanks for the updates.

PR #1073, opened 2020-10-23
Update setup.py

maajjyy:
added 3.9 version of python

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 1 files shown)


mtreinish:
Thanks for the PR, but this is a bit premature. Qiskit doesn't actually support python 3.9 yet, mostly because none of its upstream dependencies. We will bump the supported version here when all the qiskit components release as listing support for python 3.9. This will be a multi-step process as it takes time to add CI configurations running tests on python 3.9 and also we have to wait for the upstream dependencies to release with python 3.9 support before we can run anything. I'm going to close this PR now because we can't merge this without actually supporting python 3.9 and when it comes time to advertise 3.9 support we'll handle it when we bump the component versions that support 3.9 (for example see https://github.com/Qiskit/qiskit/pull/802 which added 3.8 support to the package metadata). Please feel free to reopen this if you think I'm missing something.

Issue #1052, opened 2020-10-04
Dead link in "Getting Started with Qiskit"


daniel-miller-quantum:
### Informations

- **Qiskit version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
A link is not working.

### Steps to reproduce the problem
- Go on https://qiskit.org/documentation/getting_started.html
- Search for "various forms"
- click on the link

### What is the expected behavior?
- You will be directed to a page saying "Page Not Found - Sorry, but the page you were trying to view does not exist."

### Suggested solutions
- Update/Remove the hyperlink

daniel-miller-quantum:
Sorry, this is not a bug but a documentation issue.

wmurphy-collabstar:
Hello, I'm interested in this issue. Is it okay if I work on it? Also, do you know where I can look to figure out what the link should be?

shaashwat:
Hey, I didn't see this fixed and it has been 20 days so can I go ahead and fix it?

PR #1080, opened 2020-10-27
Removed dead link in getting_started.rst

shaashwat:
### Summary
Fixed issue #1052 

### Details and comments
Removed a dead link which pointed to a non-existent page.
To recreate this issue:
1) Go on https://qiskit.org/documentation/getting_started.html
2) Search for "various forms"
3) click on the link

I fixed this issue by removing the dead link in getting_started.rst, and replacing it with plain text.

I then replaced the `QuantumCircuit.draw()` with a :meth: role pointing to the link `qiskit.circuit.QuantumCircuit.draw` . Once a make HTML is performed, this should make a direct link to the QuantumCircuit.draw() method.


docs/getting_started.rst
```diff
@@ -152,8 +152,7 @@ Step 4 : Visualize the Circuit
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can use ``QuantumCircuit.draw()`` to view the circuit that you have designed
-in the `various forms <api/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit.draw>`__
-used in many textbooks and research articles.
+in the various forms used in many textbooks and research articles.
```
SooluThomas, line 155: Hi @shaashwat, Thanks for fixing this. I think adding a link to https://qiskit.org/documentation/stubs/qiskit.visualization.circuit_drawer.html#qiskit.visualization.circuit_drawer instead would be a better replacement.

mtreinish: If you're linking to the draw method api docs do not link using external http links from qiskit.org/documentation this will break builds of the docs outside of the hosted documentation in qiskit.org. Instead use an internal reference: https://www.sphinx-doc.org/en/3.x/usage/restructuredtext/roles.html#ref-role 

Although honestly in this case I would probably just add a link above at

```
QuantumCircuit.draw()
```
and replace it with
```
:meth:`qiskit.circuit.QuantumCircuit.draw`
```

shaashwat:
@mtreinish 
I replaced the ``QuantumCircuit.draw()`` with a :meth: role pointing to the link qiskit.circuit.QuantumCircuit.draw. Once a make is performed, this should make a direct link to the QuantumCircuit.draw() method.

mtreinish (approved):

PR #1088, opened 2020-11-05
Compress docs and tutorials artifacts before upload

mtreinish:
### Summary

This commit compresses the docs build artifacts before we upload it. The
output from the docs and tutorials jobs are quite large both in space
and number of files. This causes the upload job to bog down and take a
long time while it iterates over each file. By compressing it prior to
upload it should decrease this overhead since there is a single to
upload.

### Details and comments


nonhermitian (approved):
500 -> 90 and downloads in a few seconds. !!!

PR #1096, opened 2020-11-07
Add sphinx-panels to docs

nonhermitian:
### Summary
Adds `sphinx-panels` to the docs build to support changes in Terra (https://github.com/Qiskit/qiskit-terra/pull/5354) and future updates.


nonhermitian:
For some reason i do not see the artifacts, but i only made the color darker so this should be good like last one was.

nonhermitian:
Oh, there they are looks good. pushing

PR #1103, opened 2020-11-09
Pin docplex version in CI

mtreinish:
### Summary

The latest docplex version which was released recently breaks the aqua
quadratic_program module. This commit pins the docplex version to
unblock CI, while we wait for the in-progress aqua PR Qiskit/qiskit-aqua#1420
fixing this to be included in a release.

### Details and comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,8 +80,8 @@ jobs:
         run: |
           set -e
           git clone https://github.com/Qiskit/qiskit-tutorials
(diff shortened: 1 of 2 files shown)


nonhermitian:
Perhaps pin to the previous version < 2.16

mtreinish:
Yeah, that's a better idea. I typically like to pin in constraints to an exact version to make it more deterministic, but in this case that makes sense.

t-imamichi:
https://github.com/Qiskit/qiskit-aqua/pull/1420 is merged.

PR #1104, opened 2020-11-09
Adds a (really) quick start guide to understanding QC to the docs

nonhermitian:
### Summary
Gives a really quick into to QC that can be included in the docs.  The point is not to be another set of learning materials, but rather to give a flavor of what is going on, and then point the interested reader to other content.


nonhermitian:
Looks good.

PR #1112, opened 2020-11-12
Remove autosummary templates

nonhermitian:
### Summary
The autosummary templates that were used with the old theme are causing trouble with the new one.  Removing them seems to resolve many issues.


nonhermitian:
This looks good.

PR #1139, opened 2021-01-11
Pin scipy version in CI

mtreinish:
### Summary

The recent release of scipy 1.6.0 is causing the qsvm tutorials to fail
(see Qiskit/qiskit-aqua#1498 for more details). This is blocking the
docs deploy CI job because it runs the tutorials (as well as the likely
the tutorials job too) and preventing us from uploading new versions of
the built documentation. This commit pins the scipy version to a known
working version (1.5.4) to unblock CI until we have a fix for the
tutorial failure with scipy 1.6.0.

### Details and comments

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -2,3 +2,4 @@ astroid==2.3.3
 pylint==2.4.4
 cryptography==2.5.0
 docplex==2.15.194
+scipy==1.5.4


SooluThomas (approved):
Thank you! LGTM

PR #1145, opened 2021-01-21
pin sphinx to v3.3 to fix search

nonhermitian:
### Summary
Pins Sphinx to v3.3 as v3.4 seems to break our search functionality.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 numpy>=1.17
-Sphinx>=3
+Sphinx==3.3
 qiskit-sphinx-theme>=1.7.0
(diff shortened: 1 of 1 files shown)


mtreinish:
Hmm, I thought we pinned sphinx to <3.1.0, but I just checked and that is only on terra and the other elements jobs because they emit warnings with newer sphinx versions (and the CI jobs mostly set `-W`).

mtreinish (approved):
LGTM, but is there an upstream bug we can track for this so we know when to remove the pin?

nonhermitian:
I do not know if one exists that covers this. I am thinking it just cannot find the JS anymore, but haven’t locked that down yet.

PR #1164, opened 2021-02-08
Mark versionutils extension as not parallel_read_safe

mtreinish:
### Summary

The versionutils sphinx extension we use during the combined
documentation is raising a warning on each sphinx build about it not
specifying a parallel_read_safe option. This warning while not fatal has
a potential future issue if/when we're able to set `-W` during builds this
would be a blocker. This commit returns the `parallel_read_safe` option
as False since it turns out the extension is not parallel read safe.


mtreinish:
I'm just going to close this for now. I need to dig into why the extension is not parallel read safe (it failed with it set). Setting it to false doesn't get rid of the warnings, it just changes it to a warning that parallel_read_safe is not set and it's defaulting to False to a a warning that parallel_read_safe is set to False in a parallel build and that the extension will be read serially.

PR #1174, opened 2021-02-27
Fixes #1008 by downloading the Qiskit Metapackage into a tmp folder

### LINKED ISSUE #1008: Version history table in release notes is not generated in translated docs

English Qiskit Docs: https://qiskit.org/documentation/release_notes.html#id187
Korean Qiskit Docs: https://qiskit.org/documentation/locale/ko_KR/release_notes.html#id155

We probably might want to generate the version history table in all docs.

Morcu:
…nd getting the tags from that folder

### Summary

Fixes #1008

### Details and comments

`git tag` for translated docs returns an empty list in "**Qiskit Metapackage Version**" because translated docs are in  `qiskit-community/qiskit-translations`. Instead of doing that, a new function has been aded, It downloads the git repo into a temporary folder and then the `git tag` command is executed so it always return the latest version list from the official repo.


AUTHORS
```diff
@@ -368,3 +368,4 @@ willhbang
 yang.luh
 yotamvakninibm
 Mantas Čepulkovskis
+David Morcuende Cantador
```
mtreinish, line 371: FWIW, you don't need to manually add your name here, this file is auto generated with: https://github.com/Qiskit/qiskit/blob/master/tools/generate_authors.py which gets run by the release bot every time there is a new qiskit release.

Also, note on closer inspection that when the file gets automatically updated this line will be removed and the author string from the git commit message will be used (which is `Morcu`). If you want to fix this you can either update the .mailmap file in this repo to set your preferred name or update the commit locally with:
`git config user.name "David Morcuende Cantador"` and then amend the commit with `--reset-author` locally and force push it to your fork branch.

mtreinish (commented):
The code lgtm thanks for fixing this but I before I approved and merged this I wanted to give you a chance to fix the author in the commit because the addition you made to the AUTHORS file won't persist.

AUTHORS
Morcu: Oh I didn't know, thanks for the info! It should be fixed.

mtreinish (approved):

PR #1181, opened 2021-03-04
Fix duplicate targets

jyu00:
### Summary
Rename the `.. _Release Notes_0.12.0_xx` targets so they don't collide with Terra 0.12.0.


mtreinish:
Oh, sorry I missed this PR. I already integrated this change into the bump meta release notes when I added the aer release notes. I'm going to close this since it's already done.

PR #1186, opened 2021-03-07
Remove benchmarks that measure BasicAer as transpile target

mtreinish:
### Summary

This commmit removes all the benchmarks which were using the basic aer
simulator as a target for transpilation. These benchmarks were
originally added in an attempt to show the overhead of using routing and
layout in a transpile but never did a good job of this. We have better
microbenchmarks now of each individual transpiler pass and the basic aer
target isn't a realistic use case for anything anymore. By removing
these duplicate benchmarks we get back some of our time budget that we
can use for adding new benchmarks.

There are 2 benchmarks which aren't updated to remove the basic aer
usage. The first is the quantum_volume module which is addressed in
PR #1185 as part of a larger refactor of that benchmark and the state
tomography benchmarks which are are actually running a simulation in
basic aer and its sill necessary there.

### DIFF
--- a/test/benchmarks/isometry.py
+++ b/test/benchmarks/isometry.py
@@ -36,17 +36,6 @@ class IsometryTranspileBench:
         qc.iso(iso, q[:m], q[m:])
         self.circuit = qc
 
-    def time_simulator_transpile(self, *unused):
-        transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'],
-                  seed_transpiler=0)
-
-    def track_cnot_counts(self, *unused):
-        circuit = transpile(self.circuit, basis_gates=['u1', 'u3', 'u2', 'cx'],
-                            seed_transpiler=0)
-        counts = circuit.count_ops()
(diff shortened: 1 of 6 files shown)


kdk (approved):
Looks good. It would be good as a follow up identify some use cases of transpiling for a simulator and add targeted benchmarks there for transpiler to Aer.

mtreinish:
> Looks good. It would be good as a follow up identify some use cases of transpiling for a simulator and add targeted benchmarks there for transpiler to Aer.

Yeah, I agree. My bigger issue that I was fixing in this PR was most of the tests being removed were basically just duplicating everything to run a transpile with and without a coupling map, because basic aer has the old ibmq basis set (and all the ibmq target benchmarks are still using that) which was wasting a lot of run time.

For the follow up I think we should probably run an rb and maybe a quantum volume example on aer (so we get the larger basis set). I also think in that same PR we should update the state tomography examples to run with aer instead of basic aer (since we'll be adding it to benchmark requirements).

PR #1187, opened 2021-03-09
Add benchmarks using Qiskit aer as a target

mtreinish:
### Summary

In a follow up to #1186 where we removed all the benchmarks using
BasicAer as a compilation target this commit adds some new benchmarks
that use qiskit-aer as a compilation target. Mainly a single qv
circuit (at all 4 optimization levels) and randomized benchmarking (
both single process and parallel). The state tomography benchmarks were
also updated to use aer instead of basic aer, however those benchmarks
also run a simulation so it's more than just a compilation target
difference. This use of aer is done using a pinned version for stability
in the results when we change the aer version the benchmarks using it
will have to be updated too so we know the results are associated with
a single aer version.

Since we had to invalidate the results/bump the version of the rb
benchmarks this was a good opportunity to update the ignis version to
the latest release (which has a similar pinned version arrangement).

### DIFF
--- a/test/benchmarks/transpiler_levels.py
+++ b/test/benchmarks/transpiler_levels.py
@@ -19,6 +19,7 @@ import os
 
 from qiskit.compiler import transpile
 from qiskit import QuantumCircuit
+from qiskit import Aer
 
 from .backends.fake_melbourne import FakeMelbourne
 from .utils import build_qv_model_circuit
@@ -155,6 +156,7 @@ class TranspilerLevelBenchmarks:
         large_qasm_path = os.path.join(self.qasm_path, 'test_eoh_qasm.qasm')
         self.large_qasm = QuantumCircuit.from_qasm_file(large_qasm_path)
         self.melbourne = FakeMelbourne()
+        self.aer_backend = Aer.get_backend('qasm_simulator')
 
     def time_quantum_volume_transpile_50_x_20(self, transpiler_level):
         transpile(self.qv_50_x_20, basis_gates=self.basis_gates,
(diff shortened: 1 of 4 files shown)


mtreinish:
The one thing that I'm wondering is if we want to wait until Qiskit/qiskit-terra#5868 is resolved before merging this so we can track whether the regression being tracked there is fixed (this will invalidate the stored results for previous versions of the benchmarks).

test/benchmarks/transpiler_levels.py
```diff
@@ -194,6 +196,10 @@ def time_transpile_qv_14_x_14(self, transpiler_level):
         transpile(self.qv_14_x_14, self.melbourne, seed_transpiler=0,
                   optimization_level=transpiler_level)
 
+    def time_transpile_aer_qv_14_x_14(self, transpiler_level):
+        transpile(self.qv_14_x_14, self.aer_backend, seed_transpiler=0,
+                  optimization_level=transpiler_level)
+
```
mtreinish: I need to split this out as a separate class so it doesn't interfere with the existing benchmarks' results.

mtreinish: Fixed in: https://github.com/Qiskit/qiskit/pull/1187/commits/44dc082161356be8622db5ee3a9b29e87b1bbe53

kdk (commented):
> The one thing that I'm wondering is if we want to wait until [Qiskit/qiskit-terra#5868](https://github.com/Qiskit/qiskit-terra/issues/5868) is resolved before merging this so we can track whether the regression being tracked there is fixed (this will invalidate the stored results for previous versions of the benchmarks).

Agree. Changes here look good, but on hold until the remaining regressions are resolved. Out of curiosity, does asv still show data for removed or outdated benchmarks in its UI, or do they disappear?

mtreinish:
> Agree. Changes here look good, but on hold until the remaining regressions are resolved. Out of curiosity, does asv still show data for removed or outdated benchmarks in its UI, or do they disappear?

I think they'll disappear for the deleted benchmarks so the sim ones removed in #1186 (at least that's what happened when we removed the memory benchmarks). The webui will not show results from old versions when the version string changes, so for rb and state tomography the existing data will disappear too. I think it's likely still stored in the results json (I can check), but it doesn't get displayed anymore.

HuangJunye:
@mtreinish @kdk Should this PR be merged or still on hold?

PR #1200, opened 2021-03-24
New documentation layout

nonhermitian:


nonhermitian:
I believe this is ready to go as the artifacts look good.

nonhermitian:
With the apps out this can be released.  Moreover this is needed as the apps are using the latest version of the Sphinx theme and are pointing to pages that are missing until this PR merges.  As such merging now.  Artifacts look like they should.

PR #1210, opened 2021-04-02
Put source installation instructions in new tab

nonhermitian:


mtreinish (approved):
I think there is some cleanup needed to the actual install from source contents but we can do that in a follow up, for now I think making it a tab makes sense.

PR #1221, opened 2021-04-05
Add Sphinx 3.3 to constraints

nonhermitian:
### Summary
The tutorials build was using Sphinxc `3.5.3` and we know what bad stuff happens when anything later than `3.3` is used.  As such add to constraints and hope for the best.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -4,3 +4,4 @@ cryptography==2.5.0
 docplex==2.15.194
 scipy==1.5.4
 decorator==4.4.2
+Sphinx==3.3


nonhermitian:
I think this fixed it as the tutorials build now works.

mtreinish:
We should have already been pinned to 3.3: https://github.com/Qiskit/qiskit/blob/master/requirements-dev.txt#L2 (which you pushed in #1145). If you look at that the logs from the last deploy job build: https://github.com/Qiskit/qiskit/runs/2270822496?check_suite_focus=true#step:5:56 it installed Sphinx==3.3.0

nonhermitian:
I saw this and assumed it was not holding:

https://github.com/Qiskit/qiskit/runs/2269938796#step:5:15

```
 Collecting sphinx
 Downloading Sphinx-3.5.3-py3-none-any.whl (2.8 MB)
```

PR #1222, opened 2021-04-05
[DO NOT MERGE] Revert "More Qiskit 0.25.0 release note fixes"

nonhermitian:
Looking to see if this has any effect on the tutorials builds.  I am assuming not, but does not hurt to give it a shot.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -187,7 +187,7 @@ New Features
 
 - The :class:`~qiskit.quantum_info.Statevector` and
   :class:`~qiskit.quantum_info.DensityMatrix` classes now have
-  :meth:`~qiskit.quantum_info.Statevector.draw` methods which allow objects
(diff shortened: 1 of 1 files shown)


nonhermitian:
@mtreinish take a look at the tutorials docs in this PR.  It seems they are fine.

mtreinish:
The PR CI output isn't fully indicative of what the build output for the publish job will look like. This is because we set `QISKIT_DOCS_BUILD_TUTORIALS='always'` when we build for publishing, but we do not set this for the PR CI docs jobs (in the interest of time). All the CI builds will look like they work, because I believe this is an issue with an incompatibility between the latest docutils and nbsphinx.

After temporarily disabling `QISKIT_DOCS_BUILD_TUTORIALS` the published docs are working now, I'm gonna try to come up with a workaround for the issues when building with it set now.

nonhermitian:
Closing as issue found.

PR #1227, opened 2021-04-07
Add additional pages to frontmatter

nonhermitian:
Puts the intro to QC and intro to Qiskit in the frontmatter (they are still accessible via in content links as well).


nonhermitian:
Artifacts look good.

PR #1232, opened 2021-04-14
Add layout and variables to docs

nonhermitian:
Adds layout and variables to the local docs so that changes can be made here rather than in the main Sphinx theme repo.

- Also changes the nav entry from "Qiskit Providers" -> "Providers"


nonhermitian:
Artifacts are good pushing

PR #1235, opened 2021-04-15
Change DOI badge and bibtex reference to point to the latest version

1ucian0:
The DOI badge is pointing to an old version of Qiskit instead of the `latestdoi`.


mtreinish:
Do you think we should update doi reference in the generate bibtex script too?

1ucian0:
yeap... it should. Done in 6b93aa390

mtreinish:
> yeap... it should. Done in [6b93aa3](https://github.com/Qiskit/qiskit/commit/6b93aa39013c25b6b214dc75de25fb326f0de543)

Actually you shouldn't do it just in there, that file is autogenerated at release time (and will be overwritten today by 0.25.1). You need to update the generator script: https://github.com/Qiskit/qiskit/blob/master/tools/generate_bibtex.py#L12

1ucian0:
done

mtreinish (approved):

PR #1236, opened 2021-04-15
README: breaking aqua into the application components

1ucian0:
Not sure if it is too soon for this. This PR adds the application modules to the README file. Probably discussion is needed on the details (like, where to put Aqua).


woodsp-ibm:
I guess this has been sat here for a while. Maybe the application modules could be split into a separate list since they require extra to install them, along with perhaps the info beyond the basic `pip install qiskit` that gets them installed.

mtreinish:
I'm going to close this as superseded by #1323, feel free to reopen if you feel that's incorrect or there is more to discuss here

PR #1249, opened 2021-05-05
Fix inherited members showing up in subclasses

nonhermitian:
### Summary
from: https://github.com/Qiskit-Partners/qiskit-runtime/pull/5

We have had a long standing issue with subclasses showing their inherited members even though we say not to.  This causes a huge increase in docs build time because every circuit library item is a subclass of `QuantumCircuit` and the docs tries to build the same methods over and over again because of this.

This gets us part of the way there by only listing those methods that are not inherited.  It still does not stop the docs from listing the details of the methods further down, but at least the methods table is free from inherited items.  I am not 100% sure why this does not work here, when it does in the runtime docs.


mtreinish (dismissed):

mtreinish:
Actually looking at this a bit more closely I'm not sure we want this or not. All this does in practice is remove the inherited methods out of the table on the top of the class doc right? It doesn't look like it has a noticeable effect on build time at least comparing 2 CI logs it's still in the 30-40min range. Do you think it's worth it to clear out the method list/table at the top of the doc page even if every method is still listed in the body?

I did look a this problem with @Cryoris a while ago and from what I remember the issue was that sphinx autodoc passes around a function pointer and when I started playing with manually tweaking the sphinx internal state there wasn't anyway I could definitively say whether a `.cx()` method was from `QuantumCircuit` or a subclass of `QuantumCircuit`. The function pointer object always said it was from QuantumCircuit (or something like that, the details are hazy as it was probably ~1 yr ago).

nonhermitian:
Yeah, I dont think it quite gives us what we need.

jakelishman:
Sphinx' default is to not show inherited members.  However, in `conf.py` of the metapackage and terra, but _not_ runtime, there's a line
```python
autodoc_default_options = {
    "inherited-members": None,
}
```
I suspect this line was added to mean "don't inherit any members", but what it really means is "add the `:inherited-members:` flag with no options to everything".  So it's like the nuclear version of "inherit" rather than the disable (which would be setting it to `False`, I believe, but that's also the default).  As the template doesn't override it, `conf.py` adds `:inherited-members:` to the `.. autoclass ::` directive, and we get into trouble.

I've got a working version of a similar template.  In mine, no inherited members are generated, and for neatness' sake, I had all the attributes put on the same page, rather than getting their own HTML file.  Even with my template only applied to classes in `circuit.library` (for now - I was just testing things), the number of stubs dropped from ~15,000 to ~5,000, and I was able to build the entire documentation in serial on one core of my laptop in 15 minutes.  Admittedly, I don't know how long it used to take before, but I'm like 90% sure it was much longer.  I can see in the output that the number of calls to `jupyter-execute` is _way_ lower.

jakelishman:
Update: I re-ran the full docs build of the current terra `main` on a single core, and it took 47 minutes.  So even with only the circuit library modified, that was a 3x speedup.

There are two main components to the time saving: reducing the number of calls to `jupyter-execute`, and reducing the total number of HTML files that have to be templated and output.  The former is solved by not generating documentation for inherited members, the latter is solved by included more stuff on the same page, as opposed to giving attributes/methods their own pages.  There's no real harm in letting methods have their own pages (the first line of the documentation goes on the class page as well), but putting attributes all on the class page is probably very sensible.

nonhermitian:
The whole sphinx build is a heavily modded pytorch theme. 

If I understand you, if we drop the conf setting your saying that takes care of a lot, the rest being too many jupyter instances, one per page.

jakelishman:
The inherited members thing is part of the slowdown, because it triggers a bunch of unnecessary calls to jupyter, but collapsing attributes down to be just listed on the class pages is another very major factor.  Qiskit/qiskit-terra#6865 implements all of this, and tidies up a lot of redundant configuration information and things like this so it should be easier to work with in the past.  That will apply to Terra's CI only right now, until we patch the metapackage in a compatible manner.

I think it's better to handle this at the moment by updating the metapackage's `conf.py` to pull in Terra's templates as well as its API docs, rather than us manually hardcoding Terra's templates into this repo, to avoid having two separate implementations that will almost certainly diverge at some point in the future.  I'm not certain, but it may be possible to use different template paths for different sections of the API docs, so the individual packages built together in the metapackage documentation can all have their own templates without needing to synchronise.

The reason for preferring having Terra's templates managed by Terra rather than the metapackage is because Terra's CI tests the documentation build, so we need to make sure it's the same as the metapackage and that it doesn't take excessively long (like it does now).

mtreinish:
> I think it's better to handle this at the moment by updating the metapackage's `conf.py` to pull in Terra's templates as well as its API docs, rather than us manually hardcoding Terra's templates into this repo, to avoid having two separate implementations that will almost certainly diverge at some point in the future. I'm not certain, but it may be possible to use different template paths for different sections of the API docs, so the individual packages built together in the metapackage documentation can all have their own templates without needing to synchronise.
> 
> The reason for preferring having Terra's templates managed by Terra rather than the metapackage is because Terra's CI tests the documentation build, so we need to make sure it's the same as the metapackage and that it doesn't take excessively long (like it does now).

I think as an end state this would be fine. But maybe after we've finished the deprecation and removal of aqua, ignis, and the ibmq provider from the metapackage. At that point most of the documentation will just be terra's anyway so we don't have to worry too much about side effects. But for the short/medium term I think just keeping a copy in the metapackage might make sense for now just to avoid too much jumping around to fix hosted docs issues. But it's also not a big deal to do this now either.

jakelishman:
Ah ah, I was misunderstanding a little bit about how the metapackage pulled in the Terra docs to build them before.  The actual answer to Paul then, is that this PR didn't quite work because the template in it sets `:members:` (which is what makes autodoc generate the inline description of the methods, not the table), and `conf.py` tacks on `:inherited-members:`.  So Paul was right, and removing that line from `conf.py`, or explicitly setting `:no-inherited-members:` in the template should be all that you need.

Note that setting those values doesn't affect what's in the Jinja list `methods`; autosummary has those populated before autodoc even runs, which is why you still need the conditional printing that Paul's got here.

mtreinish:
I'm closing this as #1328 is superseding this, it implements the same thing based on what @jakelishman did in https://github.com/Qiskit/qiskit-terra/pull/6865 (as that metapackage bump includes that terra pr)

PR #1261, opened 2021-05-24
Add support for stable versions of documentation

mtreinish:
### Summary

This commit adds support for hosting stable versions of the
documentation moving forward. This is accomplished by adding a new ci
workflow on tags that will upload docs to stable/$VERSION where $VERSION
is the first 2 components ('major.minor') from the version string. This
will let us maintain a hosted copy of each minor release of
documentation for users still running on older versions. The version
template and custom sphinx extensions are updated to determine the list
of versions from the current state of the git repo.

This PR assumes that we've manually updated 0.24.x, 0.25.x, and 0.26.x
to 'documentation/stable/0.24', 'documentation/stable/0.25', and
'documentation/stable/0.26' respectively. This is a blocker for merging
this because until that done the version menu will have broken links.

### Details and comments

### DIFF
--- a/.github/workflows/docs_tag.yml
+++ b/.github/workflows/docs_tag.yml
@@ -1,4 +1,4 @@
-name: Docs Publish
+name: Stable Docs Publish
 on:
   push:
     tags:
--- a/.github/workflows/docs_tag.yml
+++ b/.github/workflows/docs_tag.yml
@@ -1,0 +1,29 @@
+name: Docs Publish
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -U virtualenv setuptools wheel tox
+        sudo apt-get install graphviz pandoc
+    - name: Build and publish
+      env:
+        encrypted_rclone_key: ${{ secrets.encrypted_rclone_key }}
+        encrypted_rclone_iv: ${{ secrets.encrypted_rclone_iv }}
+        QISKIT_DOCS_BUILD_TUTORIALS: 'always'
+      run: |
+        tools/deploy_documentation_tag.sh
--- a/docs/_templates/versions.html
+++ b/docs/_templates/versions.html
@@ -15,7 +15,7 @@
     <dl>
       <dt>Previous Releases</dt>
       {% for version in version_list %}
-        <dd><a class="version" href="/documentation/stable/{{ version }}">{{ version }}</a></dd>
+        <dd><a class="version" href="/documentation/stable/{{ version }}/index.html">{{ version }}</a></dd>
       {% endfor %}
     </dl>
   </div>
--- a/docs/_templates/versions.html
+++ b/docs/_templates/versions.html
@@ -15,7 +15,7 @@
     <dl>
       <dt>Previous Releases</dt>
       {% for version in version_list %}
-        <dd><a class="version" href"/documentation/stable/{{ version }}">{{ version }}</a></dd>
+        <dd><a class="version" href="/documentation/stable/{{ version }}">{{ version }}</a></dd>
       {% endfor %}
     </dl>
   </div>
--- a/docs/_templates/versions.html
+++ b/docs/_templates/versions.html
@@ -12,6 +12,12 @@
       {% endfor %}
     </dl>
     {% endif %}
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
+        <dd><a class="version" href"/documentation/stable/{{ version }}">{{ version }}</a></dd>
+      {% endfor %}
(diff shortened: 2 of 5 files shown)


docs/_templates/versions.html
```diff
@@ -12,6 +12,12 @@
       {% endfor %}
     </dl>
     {% endif %}
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
+        <dd><a class="version" href"/documentation/stable/{{ version }}">{{ version }}</a></dd>
```
mtreinish:
```suggestion
        <dd><a class="version" href="/documentation/stable/{{ version }}">{{ version }}</a></dd>
```

.github/workflows/docs_tag.yml
```diff
@@ -0,0 +1,29 @@
+name: Docs Publish
+on:
+  push:
+    tags:
+      - '*'
```
SooluThomas, line 5: Nice!

SooluThomas (dismissed):
LGTM!

docs/_templates/versions.html
```diff
@@ -12,6 +12,12 @@
       {% endfor %}
     </dl>
     {% endif %}
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
+        <dd><a class="version" href="/documentation/stable/{{ version }}">{{ version }}</a></dd>
+      {% endfor %}
+    </dl>
```
SooluThomas, line 20: I'm just trying to run locally because I think the versions and the language will show up in the same drop down menu. Not sure if we want it that way.

mtreinish: They will, I don't think there is anyway around this. But there will be 2 sections, one on top that will list the languages first and then a section for previous releases on the bottom.

docs/_templates/versions.html
```diff
@@ -12,6 +12,12 @@
       {% endfor %}
     </dl>
     {% endif %}
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
+        <dd><a class="version" href="/documentation/stable/{{ version }}">{{ version }}</a></dd>
```
mtreinish:
```suggestion
        <dd><a class="version" href="/documentation/stable/{{ version }}/index.html">{{ version }}</a></dd>
```

docs/_templates/versions.html
mtreinish: For an example of what the different sections look like in practice you can take a look at the retworkx documentation (only the dev version has it currently): https://qiskit.org/documentation/retworkx/dev/

.github/workflows/docs_tag.yml
```diff
@@ -0,0 +1,29 @@
+name: Docs Publish
```
SooluThomas:
```suggestion
name: Stable Docs Publish
```

tools/deploy_documentation_tag.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# Script for pushing the documentation to the qiskit.org repository.
```
SooluThomas:
```suggestion
# Script for pushing the stable documentation.
```

SooluThomas (commented):
Nitpicks

mtreinish:
I've started uploading manually built version of documentation, https://qiskit.org/documentation/stable/0.25/index.html and https://qiskit.org/documentation/stable/0.24/index.html are up now (0.26 is being built right now). So we need to merge this sooner rather than later to ensure we don't accidentally delete those.

SooluThomas (approved):

drewrisinger:
Thank you for merging this! Having old documentation helps a lot!

veenaiyuri:
Thank you for this. Is it possible to move the versioned docs to a more visible location? Perhaps similar to how PyTorch hosts their versioned docs? https://pytorch.org/docs/stable/index.html

mtreinish:
> Thank you for this. Is it possible to move the versioned docs to a more visible location? Perhaps similar to how PyTorch hosts their versioned docs? https://pytorch.org/docs/stable/index.html

I'm not sure that I follow. Right now https://pytorch.org/docs/stable/index.html is roughly equivalent to https://qiskit.org/documentation/index.html, the main qiskit documentation page is for the latest stable release (in this case we're currently on 0.26.2). This PR added hosted copies for old stable releases such as https://qiskit.org/documentation/stable/0.24/index.html (which is equivalent to something like https://pytorch.org/docs/1.7.1/ in the pytorch docs). Are you asking to move `documentation/stable/$version` to `documentation/$version` or something else? If it's about how we expose the links to the old versions we're kind of stuck with how the qiskit_sphinx_theme package works right now, but we might be able to come up with something else. Can you open a new issue with the details of what you're looking for and we can discuss it in there.

veenaiyuri:
Yes, it is about how we expose the links to the old versions. Currently, the links are buried in the language switcher drop-down and most people will not notice that the previous releases are there. I will open a new issue with details. Thanks.

PR #1268, opened 2021-06-10
Add redirects for application tutorials

mtreinish:
### Summary

The tutorials for the application modules have been moved out of the
qiskit-tutorials repository to the application module documentation.
This commit adds the approriate redirects so existing links work as
expected. It also updates the tutorials index and requirements list
to remove things that were only there for the application tutorials.

### Details and comments

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -76,12 +76,71 @@ extensions = [
     'sphinx_reredirects'
 ]
 
+optimization_tutorials = [
+    '1_quadratic_program',
+    '2_converters_for_quadratic_programs',
+    '3_minimum_eigen_optimizer',
+    '4_grover_optimizer',
+    '5_admm_optimizer',
+    '6_examples_max_cut_and_tsp',
+    '7_examples_vehicle_routing',
+    '8_cvar_optimization'
+]
+
+finance_tutorials = [
+    '01_portfolio_optimization',
+    '02_portfolio_diversification',
+    '03_european_call_option_pricing',
+    '04_european_put_option_pricing',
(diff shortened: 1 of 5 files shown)


manoelmarques:
It would be better not to list each tutorial and instead just point to the index. We are constantly updating tutorials and may want to upload new ones in between releases and it would constrain us if we have to release Qiskit docs as well.

docs/conf.py
```diff
@@ -76,13 +76,72 @@
...
+optimization_tutorials = [
+    '1_quadratic_program',
+    '2_converters_for_quadratic_programs',
+    '3_minimum_eigen_optimizer',
+    '4_grover_optimizer',
+    '5_admm_optimizer',
+    '6_examples_max_cut_and_tsp',
+    '7_examples_vehicle_routing',
+    '8_cvar_optimization'
+]
```
manoelmarques, line 89: It would be better not to list each tutorial and instead just point to the index. We are constantly updating tutorials and may want to upload new ones in between releases and it would constrain us if we have to release Qiskit docs as well.

mtreinish: I've updated this in: https://github.com/Qiskit/qiskit/pull/1268/commits/a735146c18448e63ae01ef0d547609a6dcba5409 we still need to list out the individual tutorials on the old side of the redirect so the plugin knows how the source url. If I try to wildcard it it fails because there are no documents off `tutorials/optimization/` for the wild card to match (this was the first thing I tried).

SooluThomas (approved):
LGTM

PR #1280, opened 2021-06-28
Add quantum_info benchmarks - Clifford & Cnot-Dihedral

eliarbel:

### DIFF
--- a/test/benchmarks/quantum_info.py
+++ b/test/benchmarks/quantum_info.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=missing-docstring,invalid-name,no-member
+# pylint: disable=attribute-defined-outside-init
+
+from qiskit.quantum_info import random_clifford, Clifford, decompose_clifford
+from qiskit.quantum_info import random_cnotdihedral, CNOTDihedral
+import numpy as np
+
+
+class RandomCliffordBench:
+    params = ['1,3000', '2,2500', '3,2000', '4,1500', '5,1000', '6,700']
+    param_names = ['nqubits,length']
+    timeout = 300
+
+    def time_random_clifford(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        [random_clifford(nqubits) for _ in range(length)]
+
+class CliffordComposeBench:
+    params = ['1,7000', '2,5000', '3,5000', '4,2500', '5,2000']
+    param_names = ['nqubits,length']
+    timeout = 300
+
+    def setup(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        self.random_clifford = [random_clifford(nqubits) for _ in range(length)]
+
+    def time_compose(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        clifford = Clifford(np.eye(2 * nqubits))
+        [clifford.compose(self.random_clifford[i]) for i in range(length)]
+
+class CliffordDecomposeBench:
+    params = ['1,1000', '2,500', '3,100', '4,50', '5,10']
+    param_names = ['nqubits,length']
+    timeout = 300
+
+    def setup(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        self.random_clifford = [random_clifford(nqubits) for _ in range(length)]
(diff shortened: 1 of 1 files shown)


test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,92 @@
...
+
+from qiskit.quantum_info import random_clifford, Clifford, decompose_clifford
+from qiskit.quantum_info import random_cnotdihedral, CNOTDihedral
+import numpy as np
+
+
+class RandomCliffordBench:
+    params = ['1,3000', '2,2500', '3,2000', '4,1500', '5,1000', '6,700']
+    param_names = ['nqubits,length']
+    timeout = 300
```
mtreinish: Do you have a sense for how long these benchmarks take in total? If it's 5min per iteration this class alone would add up to 1+ hrs for each commit which is probably too much. So I'm trying to get a feel for how many of these we can add without removing other benchmarks to offset the increased time.

eliarbel: I tried to keep it quick (but not too much, so as to have a better sense of timing). So I tuned each iteration to take at most 1 second (on my i7 gen8 laptop). Overall it takes ~30 seconds running everything  with `asv dev` (i.e. I haven't timed end-to-end with the asv env creation part). 
Is that reasonable or do you want me to  reduce the iterations further  ?

mtreinish:
```suggestion
```

test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,92 @@
...
+    def time_random_clifford(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        for _ in range(length):
+            random_clifford(nqubits)
+
+
+class CliffordComposeBench:
+    params = ['1,7000', '2,5000', '3,5000', '4,2500', '5,2000']
+    param_names = ['nqubits,length']
+    timeout = 300
```
mtreinish:
```suggestion
```

test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,92 @@
...
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        clifford = Clifford(np.eye(2 * nqubits))
+        for i in range(length):
+            clifford.compose(self.random_clifford[i])
+
+
+class CliffordDecomposeBench:
+    params = ['1,1000', '2,500', '3,100', '4,50', '5,10']
+    param_names = ['nqubits,length']
+    timeout = 300
```
mtreinish:
```suggestion
```

test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,92 @@
...
+    def time_decompose(self, nqubits_length):
+        length = int(nqubits_length.split(',')[1])
+        for i in range(length):
+            decompose_clifford(self.random_clifford[i])
+
+
+class RandomCnotDihedralBench:
+    params = ['1,2000', '2,1500', '3,1200', '4,1000', '5,800', '6,700']
+    param_names = ['nqubits,length']
+    timeout = 300
```
mtreinish:
```suggestion
```

test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,92 @@
...
+    def time_random_cnotdihedral(self, nqubits_length):
+        (nqubits, length) = map(int, nqubits_length.split(','))
+        for _ in range(length):
+            random_cnotdihedral(nqubits)
+
+
+class CnotDihedralComposeBench:
+    params = ['1,1500', '2,400', '3,100', '4,40', '5,10']
+    param_names = ['nqubits,length']
+    timeout = 300
```
mtreinish:
```suggestion
```

test/benchmarks/quantum_info.py
mtreinish: Nah this is fine, I ran it locally and including venv and build time it took ~4min to run everything here. Which we won't really notice in our automated runs.

mtreinish (commented):
Overall LGTM, just one small inline comment. We shouldn't override the default timeout unless we actually need it. These tests take way less than 60 sec per execution so we don't need to set the timeout to 5min.

eliarbel:
I removed the custom timeouts, thanks for the review.

test/benchmarks/quantum_info.py
```diff
@@ -0,0 +1,87 @@
...
+# pylint: disable=attribute-defined-outside-init
+
+from qiskit.quantum_info import random_clifford, Clifford, decompose_clifford
+from qiskit.quantum_info import random_cnotdihedral, CNOTDihedral
+import numpy as np
+
+
+class RandomCliffordBench:
+    params = ['1,3000', '2,2500', '3,2000', '4,1500', '5,1000', '6,700']
+    param_names = ['nqubits,length']
```
mtreinish, line 25: So I missed this before. I think doing it as a string is fine, but the normal asv pattern for this would be something like:

```suggestion
    params = [[1, 2, 3, 4, 5, 6], [3000, 2500, 2000, 1500, 1000, 700]
    param_names = ['nqubits', 'length']
```

and then in the setup method raise a `NotImplementedError` for the combinations you don't want to run. But that gets messy since there is only a single pairing so the approach you did with the string split is nicer.

mtreinish (approved):

PR #1300, opened 2021-07-13
Add release date to version history table

mtreinish:
### Summary

This commit modifies the custom version table sphinx extension to add a
new column with the release date. It just uses the date from the git log
of the commit which was tagged. While in theory this could lead to a
mismatch if we tagged an old commit (or there was a significant delay
between commit and tagging) in practice the tags are pushed within
minutes of whenever the tagged commit merges so there will be no
difference, and using the log makes the parsing of the output a bit
simpler.

### Details and comments

A preview of this output from local build is:

Screenshot_2021-07-13_16-01-30


SooluThomas (approved):
LGTM

PR #1305, opened 2021-07-29
Add back versioning strategy docs

mtreinish:
### Summary

Prior to #1200 there was a useful section in the development strategy
page on how the qiskit metapackage version is determined and the rules
around when versions get bumped and how packages are tracked in the
metapackage. But this was deleted along with the entire document as part
of #1200. This commit adds back that section and expands it to include
mention of optional qiskit extras (which is mainly the application
modules and soon to be qiskit-experiments) to the end of the
contributing guide.


SooluThomas (approved):
LGTM

PR #1316, opened 2021-08-03
Fix bash syntax is docs deploy tag script

mtreinish:
### Summary

This commit fixes some bash syntax issues in the docs deploy tag script.
These were preventing the script from running successfully as the version
number was not being constructed correctly and the script would error
when calling sphinx-build. This should fix the jobs so we no longer need
to manually upload stable documentation at release time.

### DIFF
--- a/tools/deploy_documentation_tag.sh
+++ b/tools/deploy_documentation_tag.sh
@@ -26,12 +26,12 @@ pwd
 CURRENT_TAG=`git describe --abbrev=0`
 IFS='.'
 read -ra VERSION <<< "$CURRENT_TAG"
-STABLE_VERSION="${VERSION[0]}\.${VERSION[1]}"
+STABLE_VERSION=${VERSION[0]}.${VERSION[1]}
 
 # Build the documentation.
-tox -edocs -- -D content_prefix=documentation/stable/$STABLE_VERSION -j auto
+tox -edocs -- -D content_prefix=documentation/stable/"$STABLE_VERSION" -j auto
 
 # Push to qiskit.org website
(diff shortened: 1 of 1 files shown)


SooluThomas (dismissed):
LGTM! Thank you for fixing this

tools/deploy_documentation_tag.sh
```diff
@@ -26,12 +26,12 @@ pwd
 CURRENT_TAG=`git describe --abbrev=0`
 IFS='.'
 read -ra VERSION <<< "$CURRENT_TAG"
-STABLE_VERSION="${VERSION[0]}\.${VERSION[1]}"
+STABLE_VERSION=${VERSION[0]}.${VERSION[1]}
```
SooluThomas: I get stable version as `0.xx.0.` when I try this line locally. 

This is what I used for building the [translated ML docs](https://github.com/qiskit-community/qiskit-translations/blob/master/tools/deploy_ML_translated_documentation.sh#L23)
```suggestion
STABLE_VERSION =`echo $VERSION | cut -d "." -f -2`
```

SooluThomas (approved):

jakelishman:
I don't imagine it matters much in these cases here, but the `` `command` `` syntax in bash is semi-deprecated in favour of the POSIX-compliant `$(command)`-style substitution.

mtreinish:
Yeah, I don't think it's not a big issue for this. If ubuntu ships a bash version in the future where that syntax actively becomes a problem we can fix it then. But I expect a bunch of our scripts will have the same issue so it'll be a big thing we'll have to fix when the CI VM images update

PR #1317, opened 2021-08-10
Adding Pauli benchamrks

eliarbel:
### Summary
Benchmarking various Pauli & PauliList related methods


eliarbel:
@chriseclectic can you please review the Pauli & PauliList benchmarks I've added ? Let me know if you think more methods should be included. 
Also, should we add SparsePauliOp benchmarks at this point or should we wait for its conversion to be PauliList-based first ?

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +87,66 @@ def time_compose(self, nqubits_length):
...
+            c1 = random_clifford(nqubits)
+            p1.evolve(c1)
+
+
+class PauliListBench:
+    def time_basic_ops(self):
+        nqubits = 500
+        length = 500
+
+        pl1 = PauliList([random_pauli(nqubits, True)
```
ikkoham: How about using `random_pauli_list`?

eliarbel: Yes, good idea, will do

eliarbel: Ikko, BTW why `random_pauli_list` is not in the import list of `quantum_info` ?

ikkoham: My bad. I'll add it soon.

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +88,95 @@ def time_compose(self, nqubits_length):
...
+    def time_basic_ops(self, nqubits):
+        iterations = 200
+        for _ in range(0, iterations):
+            p1 = random_pauli(nqubits, True)
+            p2 = random_pauli(nqubits, True)
+
+            p1.compose(p2)
+            p1.evolve(p2)  # by another Pauli, so by composition
+            p1.commutes(p2)
+            p1.to_instruction()
```
mtreinish: I know we discussed this before, but I'm still a bit concerned by the bundling of operations here. It makes it more difficult to isolate regressions. Like if this benchmark flags a regression on PR X how do we know if it's the conversion to an instruction, or composition that had the issue?

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +88,95 @@ def time_compose(self, nqubits_length):
...
         for i in range(length):
             cxdihedral.compose(self.random_cnotdihedral[i])
+
+
+class PauliBench:
+    params = [100, 200, 300, 400, 500]
+
+    def time_basic_ops(self, nqubits):
+        iterations = 200
+        for _ in range(0, iterations):
```
mtreinish: Normally we don't do the iteration like this in the benchmark and rely on asv to do multiple samples for us. You can tweak how it does this with attributes: https://asv.readthedocs.io/en/stable/benchmarks.html#timing-benchmarks

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +88,95 @@ def time_compose(self, nqubits_length):
...
+            p1 = random_pauli(nqubits, True)
+            p2 = random_pauli(nqubits, True)
```
mtreinish: This probably needs a seed for consistency between runs.

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +88,95 @@ def time_compose(self, nqubits_length):
         cxdihedral = CNOTDihedral(num_qubits=nqubits)
         for i in range(length):
             cxdihedral.compose(self.random_cnotdihedral[i])
+
+
+class PauliBench:
+    params = [100, 200, 300, 400, 500]
```
mtreinish, line 94: Can you also set a `param_names` attribute here so the graphs render this as num_qubits:

```suggestion
    params = [100, 200, 300, 400, 500]
    param_names = ["num_qubits"]
```

test/benchmarks/quantum_info.py
```diff
@@ -85,3 +88,95 @@ def time_compose(self, nqubits_length):
...
+
+    def time_basic_ops(self, nqubits):
+        iterations = 200
+        for _ in range(0, iterations):
+            p1 = random_pauli(nqubits, True)
```
mtreinish: Does it make sense to split the random pauli generation out into a `setup()` method so we leave the timed portion to just the ops and not the random generation (which honestly will probably be mostly constrained on numpy's rng and system entropy).

mtreinish (commented):
I left some comments on the first benchmark class but I think they apply to all the classes in the file

eliarbel:
Thanks @mtreinish  for the review and comments. I've addressed almost all them. The only thing I've left as before is the use of unfixed random seed. I think it's better this way as otherwise we are left with benchmarking the same samples and potentially miss statistical chance of hitting a corner case. Given a known seed for each run, is there a way to log or publish it in ASV somehow in case we need to use it for debug ?

test/benchmarks/quantum_info.py
eliarbel: Done, removed

test/benchmarks/quantum_info.py
eliarbel: That one I would leave as is, to have some statistical coverage of the possible Pauli's. Theoretically we may find some corner case

test/benchmarks/quantum_info.py
eliarbel: Those are very quick, but split I put it in `setup()` anyway.

test/benchmarks/quantum_info.py
eliarbel: no problem, I split them into distinct methods. This will entail many graphs in the dashboard. If it will bother us we can revert.

mtreinish (approved):
LGTM, this looks good to me. I'm still a bit concerned about the seed but it hopefully shouldn't cause too much noise in the data. Tracing it back to a seed value though will be tricky. The only output we have is to stdout or stderr which will only get printed during the run and not get associated with the generated graphs. So tracing an outlier back will require manually looking at the logs and pulling the seed out, so it's probably not worth the effort. If the benchmarks are too noisy after a few data points we can always go in and change things.

I also tested these all locally and they're quite quick so it won't add any noticeable delay to our per commit benchmarking time.

eliarbel:
Thanks @mtreinish !

PR #1334, opened 2021-09-16
Remove cvxpy cap

mtreinish:
### Summary

The cvxpy cap is forcing a really ancient version of cvxpy to be
installed in the  docs publishing jobs causing the tutorials to fail.
I'm not quite sure why it's just happening today while last week
things worked fine. I expect it is another package's requirements
confusing pip's dependency solver. But either way this cap is no
longer needed as the tutorials ci test job runs fine with cvxpy
unconstrained.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -14,7 +14,7 @@ ipywidgets>=7.3.
 pillow>=4.2.1
 seaborn>=0.9.0
 nbsphinx
-cvxpy<1.1.0
+cvxpy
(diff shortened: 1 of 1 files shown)


mtreinish:
It looks like the correct cvxpy version is getting installed now: https://github.com/Qiskit/qiskit/pull/1334/checks?check_run_id=3622736959#step:6:10

Hopefully this translates to the publish docs job too. I'm going to merge this now because docs builds are broken and we're missing the release documentation for qiskit 0.30.0/aer 0.9.0

PR #1344, opened 2021-10-14
Document Python venv for source installs

kevinhartman:
### Summary
- Add guidance for using Python's built-in virtual environment module `venv` for source build/install.
- Warn M1-series Mac users of Aer build issue.


docs/contributing_to_qiskit.rst
```diff
@@ -676,10 +676,52 @@ steps for each element.
 Set up the Virtual Development Environment
 ==========================================
 
-.. code-block:: sh
+To ensure consistency, use virtual environments for Qiskit source builds and development.
```
mtreinish: I would probably add a bit more here about how this ensures that the development environment is kept separate from other python environments including the system's python env.

docs/contributing_to_qiskit.rst
```diff
@@ -676,10 +676,52 @@ steps for each element.
...
+
+.. note::
+
+   **M1 Mac users:** Some Qiskit dependencies may not yet be available in PyPI. Until they are,
+   Conda is recommended.
```
mtreinish, line 687: We might want to put a note in the aer section about this too that refers to https://github.com/Qiskit/qiskit-aer/issues/1286 which has instructions on building from source until we can get cross compilation working to publish arm64 wheels for aer

mtreinish (commented):
Overall LGTM, just one spot inline I think we can expand a bit more on why using venvs is useful.

docs/contributing_to_qiskit.rst
kevinhartman: Done in a9dc419b05216c92dea8cc2a3ab6a0055ea5cb70.

docs/contributing_to_qiskit.rst
kevinhartman: Done in ed83f6d.

mtreinish (approved):
LGTM, thanks for doing this and for the quick updates

PR #1346, opened 2021-10-14
Add benchmarks for missing passes

mtreinish:
### Summary

This commit adds runtime benchmarks to the passes and mapping_passes
benchmark modules to measure additional transpiler passes in isolation
that we didn't have coverage for previously. Primarily of importance
here are the basis translator and the optimize 1q decomposition passes
which are in the preset pass managers but didn't have standalone
benchmark coverage. One pass excluded here is the TemplateOptimization
pass which would have been good to add to the benchmark but it is
exceedingly slow and took well more than 5 min per iteration.

### DIFF
--- a/test/benchmarks/mapping_passes.py
+++ b/test/benchmarks/mapping_passes.py
@@ -68,6 +68,11 @@ class PassBenchmarks:
         swap.property_set['layout'] = self.layout
         swap.run(self.dag)
 
+    def time_sabre_swap(self, _, __):
+        swap = SabreSwap(self.coupling_map, seed=42)
+        swap.property_set['layout'] = self.layout
+        swap.run(self.dag)
+
     def time_basic_swap(self, _, __):
         swap = BasicSwap(self.coupling_map)
         swap.property_set['layout'] = self.layout
@@ -111,6 +116,9 @@ class PassBenchmarks:
     def time_noise_adaptive_layout(self, _, __):
         NoiseAdaptiveLayout(self.backend_props).run(self.fresh_dag)
 
+    def time_sabre_layout(self, _, __):
+        SabreLayout(self.coupling_map, seed=42).run(self.fresh_dag)
+
 
 class RoutedPassBenchmarks:
     params = ([5, 14, 20],
@@ -152,8 +160,17 @@ class RoutedPassBenchmarks:
         self.routed_dag = StochasticSwap(self.coupling_map,
                                          seed=42).run(self.dag)
 
-    def time_cxdirection(self, _, __):
+    def time_cx_direction(self, _, __):
         CXDirection(self.coupling_map).run(self.routed_dag)
 
     def time_check_cx_direction(self, _, __):
         CheckCXDirection(self.coupling_map).run(self.routed_dag)
+
+    def time_gate_direction(self, _, __):
+        GateDirection(self.coupling_map).run(self.routed_dag)
+
+    def time_check_gate_direction(self, _, __):
+        CheckGateDirection(self.coupling_map).run(self.routed_dag)
+
+    def time_check_map(self, _, __):
+        CheckMap(self.coupling_map).run(self.routed_dag)
--- a/test/benchmarks/mapping_passes.py
+++ b/test/benchmarks/mapping_passes.py
@@ -160,7 +160,7 @@ class RoutedPassBenchmarks:
         self.routed_dag = StochasticSwap(self.coupling_map,
                                          seed=42).run(self.dag)
 
-    def time_cx_direction(self, _, __):
+    def time_cxdirection(self, _, __):
         CXDirection(self.coupling_map).run(self.routed_dag)
 
     def time_check_cx_direction(self, _, __):
--- a/test/benchmarks/passes.py
+++ b/test/benchmarks/passes.py
@@ -196,3 +196,20 @@ class PassBenchmarks:
 
     def time_remove_barriers(self, _, __):
         RemoveBarriers().run(self.dag)
+
+
+class MultiQBlockPassBenchmarks:
+    params = ([5, 14, 20],
+              [1024], [1, 2, 3, 4, 5])
+
+    param_names = ['n_qubits', 'depth', 'lock_max_size']
+    timeout = 300
+
+    def setup(self, n_qubits, depth, _):
+        seed = 42
+        self.circuit = random_circuit(n_qubits, depth, measure=True,
+                                      conditional=True, reset=True, seed=seed)
+        self.dag = circuit_to_dag(self.circuit)
+
+    def time_collect_multiq_block(self, _, __, max_block_size):
+        CollectMultiQBlocks(max_block_size).run(self.dag)
--- a/test/benchmarks/passes.py
+++ b/test/benchmarks/passes.py
@@ -92,7 +92,7 @@ class UnrolledPassBenchmarks:
 class MultipleBasisPassBenchmarks:
     params = ([5, 14, 20],
               [1024],
-              [['u', 'cx'], ['rx', 'ry', 'rz', 'r', 'rxx'],
+              [['u', 'cx', 'id'], ['rx', 'ry', 'rz', 'r', 'rxx', 'id'],
                ['rz', 'x', 'sx', 'cx', 'id']])
 
     param_names = ['n_qubits', 'depth', 'basis_gates']
--- a/test/benchmarks/passes.py
+++ b/test/benchmarks/passes.py
@@ -14,9 +14,10 @@
 
 # pylint: disable=no-member,invalid-name,missing-docstring,no-name-in-module
 # pylint: disable=attribute-defined-outside-init,unsubscriptable-object
-# pylint: disable=unused-wildcard-import,wildcard-import
+# pylint: disable=unused-wildcard-import,wildcard-import,undefined-variable
 
 
+from qiskit.circuit.equivalence_library import SessionEquivalenceLibrary as SEL
 from qiskit.transpiler.passes import *
 from qiskit.converters import circuit_to_dag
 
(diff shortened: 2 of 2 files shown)


test/benchmarks/mapping_passes.py
```diff
@@ -152,8 +160,17 @@ def setup(self, n_qubits, depth):
         self.routed_dag = StochasticSwap(self.coupling_map,
                                          seed=42).run(self.dag)
 
-    def time_cxdirection(self, _, __):
+    def time_cx_direction(self, _, __):
```
mtreinish: While I don't like this typo, if we change this after 1am EDT on 10/15 we'll invalidate results from tomorrow's run, the extra data is worth more than a more consistent benchmark method name

```suggestion
    def time_cxdirection(self, _, __):
```

mtreinish:
I think we're going to have to cut back on the basis translator benchmark, it's too slow. It's been running for ~23min and still hasn't finished the first half (the cache warming unmeasured execution)

mtreinish:
Ok, I can just drop the non ibm basis from the benchmark for now to get this running nightly:

```
[100.00%] ··· ...pleBasisPassBenchmarks.time_basis_translator         6/9 failed
[100.00%] ··· ========== ======= ================================ =========
               n_qubits   depth            basis_gates                     
              ---------- ------- -------------------------------- ---------
                  5        1024            ['u', 'cx']              failed 
                  5        1024   ['rx', 'ry', 'rz', 'r', 'rxx']    failed 
                  5        1024   ['rz', 'x', 'sx', 'cx', 'id']    329±3ms 
                  14       1024            ['u', 'cx']              failed 
                  14       1024   ['rx', 'ry', 'rz', 'r', 'rxx']    failed 
                  14       1024   ['rz', 'x', 'sx', 'cx', 'id']    839±3ms 
                  20       1024            ['u', 'cx']              failed 
                  20       1024   ['rx', 'ry', 'rz', 'r', 'rxx']    failed 
                  20       1024   ['rz', 'x', 'sx', 'cx', 'id']    1.12±0s 
              ========== ======= ================================ =========
```

But, that is a huge disparity because the timeout is set to 300 sec and all the failures are timeouts. @kdk can you think of some reason it would be so much slower for those basis sets?

mtreinish:
@georgios-ts suggested the issue with the timeouts might be because the those basis were missing `id`. Testing locally adding `id` fixes this and the benchmark runs quite quickly:

```
[100.00%] ··· ...pleBasisPassBenchmarks.time_basis_translator                 ok
[100.00%] ··· ========== ======= ====================================== ============
               n_qubits   depth               basis_gates                           
              ---------- ------- -------------------------------------- ------------
                  5        1024            ['u', 'cx', 'id']             263±0.8ms  
                  5        1024   ['rx', 'ry', 'rz', 'r', 'rxx', 'id']    617±2ms   
                  5        1024      ['rz', 'x', 'sx', 'cx', 'id']       362±0.6ms  
                  14       1024            ['u', 'cx', 'id']              649±2ms   
                  14       1024   ['rx', 'ry', 'rz', 'r', 'rxx', 'id']    1.30±0s   
                  14       1024      ['rz', 'x', 'sx', 'cx', 'id']        941±2ms   
                  20       1024            ['u', 'cx', 'id']              908±10ms  
                  20       1024   ['rx', 'ry', 'rz', 'r', 'rxx', 'id']   1.73±0.01s 
                  20       1024      ['rz', 'x', 'sx', 'cx', 'id']       1.31±0.01s 
              ========== ======= ====================================== ============
```

I think this is a bug in the basis translator or the equivalence lib because it should be able to find an equivalent with the parameterized 1q gates for an id pretty easily.

mtreinish:
https://github.com/Qiskit/qiskit-terra/pull/7146 should fix the timeouts with the basis translator benchmarks

test/benchmarks/passes.py
```diff
@@ -14,9 +14,10 @@
 
 # pylint: disable=no-member,invalid-name,missing-docstring,no-name-in-module
 # pylint: disable=attribute-defined-outside-init,unsubscriptable-object
-# pylint: disable=unused-wildcard-import,wildcard-import
+# pylint: disable=unused-wildcard-import,wildcard-import,undefined-variable
```
jakelishman, line 17: What's pylint complaining about that requires this?  By name, at least, it sounds like something we shouldn't be ignoring.

test/benchmarks/passes.py
```diff
@@ -161,3 +187,29 @@ def time_remove_diagonal_gates_before_measurement(self, _, __):
...
+
+    def time_remove_barriers(self, _, __):
+        RemoveBarriers().run(self.dag)
+
+
+class MultiQBlockPassBenchmarks:
+    params = ([5, 14, 20],
+              [1024], [1, 2, 3, 4, 5])
+
+    param_names = ['n_qubits', 'depth', 'lock_max_size']
```
jakelishman: Think the last param name is wrong here.  Perhaps `max_block_size`?

jakelishman (commented):
This looks sensible to me.

kdk:
> @georgios-ts suggested the issue with the timeouts might be because the those basis were missing `id`. Testing locally adding `id` fixes this and the benchmark runs quite quickly:
> 
Agree the timeouts are likely an instance of https://github.com/Qiskit/qiskit-terra/issues/5539 .
> 
> I think this is a bug in the basis translator or the equivalence lib because it should be able to find an equivalent with the parameterized 1q gates for an id pretty easily.

I think this is a little more complicated than it appears at first glance because IBM hardware has a non-standard (though deprecated) definition of `id` (see https://github.com/Qiskit/qiskit-terra/issues/6627 and https://github.com/Qiskit/qiskit-terra/issues/6460 ). These rules had been intentionally left out to avoid a case where the transpiler might inadvertently replace an `id` with the "equivalent" `u3` (which would be twice as long).

I agree these are logically equivalences the library should have, but perhaps after `'id'` is removed from IBM's basis gates and Qiskit's definition of `IGate` is reconsidered (e.g. from https://qiskit.org/documentation/stubs/qiskit.circuit.library.IGate.html#qiskit.circuit.library.IGate 
> Identity gate corresponds to a single-qubit gate wait cycle, and should not be optimized or unrolled (it is an opaque gate).

)

test/benchmarks/passes.py
mtreinish: It's because pylint runs with the latest terra release and there are passes I'm adding benchmarks for here that are only on main. So pylint was complaining about an undefined variable for those passes.

Honestly I should just turn pylint off for the benchmarks and use flake8 and black only since it is serving a different use case than the other things being linted in this repo

mtreinish:
> I agree these are logically equivalences the library should have, but perhaps after `'id'` is removed from IBM's basis gates and Qiskit's definition of `IGate` is reconsidered (e.g. from https://qiskit.org/documentation/stubs/qiskit.circuit.library.IGate.html#qiskit.circuit.library.IGate
> 
> > Identity gate corresponds to a single-qubit gate wait cycle, and should not be optimized or unrolled (it is an opaque gate).
> 
> )

Why does this matter for Qiskit/qiskit-terra#7146 though? If IBMQ backends are using `IGate` implicitly as a single cycle delay that's unfortunate complexity but fine, it's in the basis set so the basis translator won't do anything with it. Where this would cause issues is if IBM backend's didn't report `id` as a basis gate then that relationship would be broken. I'm just not sure I see how adding id == u(0,0,0) to the equivalence lib will break the weird special case that the ibmq provider is trying to support since when would that rule be triggered the basis translator should ignore id as long as it's in the basis. I assume there is test coverage for this in terra because Qiskit/qiskit#7146 passes all the tests.

test/benchmarks/passes.py
mtreinish:
```suggestion
    param_names = ['n_qubits', 'depth', 'max_block_size']
```

test/benchmarks/passes.py
jakelishman: Ah, that makes sense.  Yeah, given the huge number of exclusions here, stopping pylint on these files is probably a good idea.

kdk:
> Where this would cause issues is if IBM backend's didn't report `id` as a basis gate then that relationship would be broken.

I think this is exactly it. Following https://github.com/Qiskit/qiskit-ibmq-provider/pull/976 (which currently warns on `id` usage and replaces with the equivalent `delay`), the IBM backends will at some point remove `id` as a `basis_gate`, at which the transpiler should fail with an unreachable basis or unknown instruction (notwithstanding https://github.com/Qiskit/qiskit-terra/issues/5539 ) rather than implement delays that are twice as long.

> equivalence lib will break the weird special case that the ibmq provider is trying to support since

Agree that this is weird and non-intuitive behavior, but it is not at the moment IBMQ provider specific. Qiskit defines `IdGate` as `delay(durationof(SXGate())` and not `UnitaryGate([[1, 0], [0, 1]])` (which I think is what most would expect, given the name). Agree this should change in the future, but without an inconsistent intermediate state.

test/benchmarks/passes.py
```diff
@@ -88,6 +89,31 @@ def time_optimize_1q(self, _, __):
         Optimize1qGates().run(self.unrolled_dag)
 
 
+class MultipleBasisPassBenchmarks:
+    params = ([5, 14, 20],
+              [1024],
+              [['u', 'cx'], ['rx', 'ry', 'rz', 'r', 'rxx'],
```
kdk:
```suggestion
              [['u', 'cx', 'id'], ['rx', 'ry', 'rz', 'r', 'rxx', 'id'],
```
To unblock this from the discussions around https://github.com/Qiskit/qiskit-terra/pull/7146 .

kdk (commented):
One comment, otherwise LGTM.

kdk (approved):

PR #1348, opened 2021-10-20
Add a benchmark of `SparsePauliOp.__add__`

t-imamichi:
### Summary

Add a benchmark of `SparsePauliOp._add` via `SparsePauliOp.__add__`.
This method is an important building block of various algorithms (I'm working on the qubit mapper of Qiskit nature).
I'm working on the performance improvement of this method. I hope the improvement will be visible.
https://github.com/Qiskit/qiskit-terra/pull/7138
https://github.com/Qiskit/qiskit-nature/pull/397

### DIFF
--- a/test/benchmarks/quantum_info.py
+++ b/test/benchmarks/quantum_info.py
@@ -182,6 +182,9 @@ class SparsePauliOpBench:
         self.p2 = SparsePauliOp(
             random_pauli_list(num_qubits=num_qubits, size=length, phase=True))
 
+    def time_add(self, _, __):
+        self.p1.add(self.p2)
+
     def time_compose(self, _, __):
         self.p1.compose(self.p2)
 
--- a/test/benchmarks/quantum_info.py
+++ b/test/benchmarks/quantum_info.py
@@ -183,7 +183,7 @@ class SparsePauliOpBench:
             random_pauli_list(num_qubits=num_qubits, size=length, phase=True))
 
     def time_add(self, _, __):
-        self.p1.add(self.p2)
+        _ = self.p1 + self.p2
 
     def time_compose(self, _, __):
         self.p1.compose(self.p2)
--- a/test/benchmarks/quantum_info.py
+++ b/test/benchmarks/quantum_info.py
@@ -182,9 +182,6 @@ class SparsePauliOpBench:
         self.p2 = SparsePauliOp(
             random_pauli_list(num_qubits=num_qubits, size=length, phase=True))
 
-    def time_add(self, _, __):
-        _ = self.p1 + self.p2
-
     def time_compose(self, _, __):
         self.p1.compose(self.p2)
 
@@ -193,6 +190,10 @@ class SparsePauliOpBench:
 
     def time_tensor(self, _, __):
         self.p1.tensor(self.p2)
+
+    def time_add(self, _, __):
+        _ = self.p1 + self.p2
+    time_add.params = [[50, 100, 150, 200], [10000]]
 
     def time_to_list(self, _, __):
         self.p1.to_list()


jakelishman (approved):
This seems like a fine and sensible addition to me, without any real risk of increasing the benchmark runtime too much, but I'll also defer to @mtreinish in case there are any other considerations with the benchmark suite that I'm not aware of.

test/benchmarks/quantum_info.py
```diff
@@ -182,6 +182,9 @@ def setup(self, num_qubits, length):
         self.p2 = SparsePauliOp(
             random_pauli_list(num_qubits=num_qubits, size=length, phase=True))
 
+    def time_add(self, _, __):
+        self.p1.add(self.p2)
```
jakelishman: Actually, I changed my mind - `SparsePauliOp` has no `add` method!  I think this should either be `self.p1 + self.p2` (which includes the `LinearOp` mixin stuff) or `self.p1._add(self.p2)`.

jakelishman (changes requested):
On second thoughts, there's actually a slight mistake in the benchmark!  I'm also thinking that this might not go high enough in the numbers of qubits and operator length right now, but that's a consideration for _all_ the timers in this benchmark.

test/benchmarks/quantum_info.py
t-imamichi: Sorry. You are right. It's my mistake. I will rewrite it as `self.p1 + self.p2`.

jakelishman (dismissed):

t-imamichi:
I tried test_add in my environment (Ryzen 2400GE, 3.2GHz). It might be not so high. Should I make `length` larger for this benchmark?
```
[ 88.46%] ··· quantum_info.SparsePauliOpBench.time_add                                                                                              ok
[ 88.46%] ··· ============ ==========
              --             length  
              ------------ ----------
               num_qubits     100    
              ============ ==========
                   50       104±20μs 
                  100       132±50μs 
                  150       123±10μs 
                  200       148±10μs 
              ============ ==========
```

t-imamichi:
If I change `length` to 10000, the time of time_add gets comparable with that of time_compose. What do you think?
```
[ 88.46%] ··· quantum_info.SparsePauliOpBench.time_add                                                                                              ok
[ 88.46%] ··· ============ =============
              --               length   
              ------------ -------------
               num_qubits      10000    
              ============ =============
                   50       3.08±0.03ms 
                  100       4.96±0.01ms 
                  150       6.74±0.03ms 
                  200       8.39±0.09ms 
              ============ =============

[ 90.38%] ··· quantum_info.SparsePauliOpBench.time_compose                                                                                          ok
[ 90.38%] ··· ============ =============
              --               length   
              ------------ -------------
               num_qubits       100     
              ============ =============
                   50       6.11±0.01ms 
                  100       10.0±0.08ms 
                  150       14.1±0.09ms 
                  200       17.5±0.09ms 
              ============ =============
```

jakelishman:
The absolute timings are probably less of a concern, because a few milliseconds here or there shouldn't have much of an impact on anything.  The timing differences I was thinking about were whether we were measuring the method's overhead (few qubits/small lengths) or large-scale performance (many qubits/large lengths) - I don't know what the benchmark suite is really meant to be all about though, so I'll wait for Matthew to comment.  I also don't know if there are any implications from adding new benchmarks, or changing the parameter lists of existing ones.

mtreinish:
We can tweak the parameters list for the class but that may invalidate results for other results in the class (because by default becnhmarks are versioned by a hash of the code, but I can't remember if the parameters are included in that hash). But you can set the params per benchmark if you want with something like:

```
time_add.params = [[[50, 100, 150, 200], [100, 1000, 10000]]
```

which let you set different values for just this new benchmark: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#parameterized-benchmarks

t-imamichi:
Thank you. I tweaked the length. The result of time_add is same as this link.
https://github.com/Qiskit/qiskit/pull/1348#issuecomment-947648437

mtreinish (approved):

PR #1349, opened 2021-10-21
Add dynamics to other-builds.txt list

mtreinish:
### Summary

The other-builds.txt list is used to maintain a list of directories to
not overwrite when we sync new docs builds to COS. These are other doc
builds hosted in the same namespace in the hosted bucket. With the
pending release of the qiskit-dynamics project it will be uploading it's
documentation to 'documentation/dynamics' and we need to add this to the
exclude list so that future docs updates for the root documentation do
not accidentally delete the dynamics documentation. This commit adds
'dynamics/**'  to the other-builds.txt list for this reason.

### Details and comments

### DIFF
--- a/tools/other-builds.txt
+++ b/tools/other-builds.txt
@@ -8,3 +8,4 @@ finance/**
 experiments/**
 retworkx/**
 stable/**
+dynamics/**


jakelishman (approved):
This looks obviously correct.  Is there a coupled PR to dynamics or qiskit_bot, or somewhere that adds in the copying?

mtreinish:
It's tied to how rclone is used here: https://github.com/Qiskit/qiskit/blob/master/tools/deploy_documentation.sh#L32. It basically the `sync` which is like `rsync --delete` which would conflict with https://github.com/Qiskit/qiskit-dynamics/blob/main/tools/deploy_documentation.sh#L32 It just hasn't come up since dynamics hasn't had a release yet (and that script only runs on tags: https://github.com/Qiskit/qiskit-dynamics/blob/main/.github/workflows/docs.yml )

jakelishman:
Ah yep thanks, I was thinking it was getting added as if it were one of the `qiskit_elements`, and getting a folder in `apidocs` by inclusion in `conf.py`, so I was expecting it needed another change somewhere.  But I see from the links that it's included by a different mechanism (wholesale copy of the tree), and that happens on tag push.

By the way, at the moment in the deployment we have `rclone` just exclude the directories in `other-builds.txt`.  Do you think it's worth making it a deployment error if there are files that are being missed?  Having a silent suppression of some copies seems like it may bite us one day, similar to how the `conf.py` copies in the `apidocs` without explicitly checking for duplicates (I don't think `copy_tree` checks that).

mtreinish:
Well the thing is we do want the `sync` behaviour generally inside a single build because the output from a sphinx build for each document should be exactly what gets uploaded and old files or files that are missing should be removed. For example, if we remove `documnetation/apidocs/aqua.html` that should get deleted when we upload the documentation. The current approach is definitely error prone, but luckily we don't add new documentation builds that often so it's pretty unlikely that we'll add a new one and miss this. (I know I just jinxed it).

I think the best solution to avoid a potential error would be to make the process additive instead and maybe upload each project's separate build to a separate bucket or namespace and then add the configuration to link it under the `qiskit.org/documentation`  namespace. The problem with that is it requires a lot more work in the cloudflare edge function to stitch everything together then and it's pretty tricky to get right.

jakelishman:
Yeah, I agree with all that.  I think I was explaining it wrong; in this case, I mean do you think it's worth adding an extra step before `rclone`, that scans through `other-builds.txt`, and ensure that this repo's documentation build hasn't produced any of the named folders.  Then `rclone` can still exclude them from the sync, safe in the knowledge that it's only excluding deletion on the remote, and not excluding any copies from this repo, which can only have been (presumably) mistakes in our process.

(It's a question mostly unrelated to this PR, though.)

mtreinish:
Ah ok, yeah that does make sense and feels like it would be a useful addition to the deployment script. It's probably pretty unlikely in practice, but it's good to have that extra check ensure that.

PR #1358, opened 2021-11-16
Fix linter errors after doc generation

garrison:
### Summary

This pull request fixes linter errors that result after building the docs locally.

### Details and comments

I have found that `tox -elint` results in errors if it is run after the documentation has been built using `tox -edocs`.  Currently, there are 7098 different `D005 No newline at end of file` errors because the generated `rst` files in `docs/stubs` are lacking trailing newlines.

I can reproduce this problem in a python 3.8 virtual environment with tox 3.24.4.

This commit fixes the errors by ignoring the `docs/stubs` directory.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ deps =
 commands =
   pycodestyle test
   pylint -rn --rcfile={toxinidir}/.pylintrc test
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Cool, thank you!

PR #1364, opened 2021-11-22
Pin matplotlib in CI

mtreinish:
### Summary

With the recent release of matplotlib 3.5.0 bloch sphere visualizations
are failing in docs builds. This has been mitigated in terra's main
branch but hasn't been included in a release (but will be fixed in the
upcoming terra 0.19.0 release). Until that is released this commit pins
things in CI to avoid using the new version to unblock ci.

### Details and comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
         run: |
           set -e
           git clone https://github.com/Qiskit/qiskit-tutorials
-          pip install -c constraints.txt -U jupyter sphinx nbsphinx networkx sphinx_rtd_theme 'matplotlib<3.3.0' qiskit-terra[visualization] cvxpy 'pyscf<1.7.4' scikit-learn
+          pip install -c constraints.txt -U jupyter sphinx nbsphinx networkx sphinx_rtd_theme 'matplotlib<3.4.0' qiskit-terra[visualization] cvxpy 'pyscf<1.7.4' scikit-learn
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         run: |
           set -e
           git clone https://github.com/Qiskit/qiskit-tutorials
-          pip install -c constraints.txt -U jupyter sphinx nbsphinx networkx sphinx_rtd_theme 'matplotlib<3.3.0' qiskit-terra[visualization] cvxpy 'pyscf<1.7.4'
+          pip install -c constraints.txt -U jupyter sphinx nbsphinx networkx sphinx_rtd_theme 'matplotlib<3.4.0' qiskit-terra[visualization] cvxpy 'pyscf<1.7.4' scikit-learn
```
jakelishman, line 83: How come `scikit-learn` needs to be added as part of this?

mtreinish: I'm not entirely sure, but the tutorials CI jobs are failing because of a missing scikit-learn: https://github.com/Qiskit/qiskit/runs/4266031152?check_suite_focus=true#step:6:119 my guess is probably a change in the tutorials

jakelishman: How odd.  Best to put it in, then!

jakelishman (approved):
I need to do the Matplotlib 3.5 fix for Terra later this week...

PR #1374, opened 2021-12-09
Expand deprecation policy

mtreinish:
### Summary

This commit expands the deprecation policy to be a bit stricter and
clarify some edge cases that have arisen in recent releases. It adds
two sub-points to the previous deprecation policy:

1. A feature can't be deprecated until the alternative has been present
   in a released version.
2. A deprecated feature can't be removed until it's been deprecated in
   more than 1 release.

These two additional subpoint should clarify some ambiguity in the
current policy especially as we've had feedback on the current approach
to deprecation and typically have > 3 months between releases.

### Details and comments


docs/contributing_to_qiskit.rst
```diff
@@ -274,6 +274,15 @@ path. A migration path might be "stop using that feature", but in such cases
...
 is to users, in order to determine a reasonable obsolescence date.
 
+2a. The migration path must have existed in a least a prior release before the
+new feature can be deprecated. For example, if you have a function ``foo()``
+which is going to be replaced with ``bar()`` you can'd deprecate the ``foo()``
+function in the same release that introduces ``bar()``. The ``bar()`` function
+needs to be available in a release prior to the deprecation of ``foo()``. This
+is necessary to enable downstream consumers of Qiskit that maintain their
+own libraries to write code that works with > 1 release at a time which is
+important for the entire ecosystem.
```
jakelishman:
```suggestion
which is going to be replaced with ``bar()`` you can't deprecate the ``foo()``
function in the same release that introduces ``bar()``. The ``bar()`` function
needs to be available in a release prior to the deprecation of ``foo()``. This
is necessary to enable downstream consumers of Qiskit that maintain their
own libraries to write code that works with > 1 release at a time, which is
important for the entire ecosystem.
```

docs/contributing_to_qiskit.rst
```diff
@@ -282,14 +291,24 @@ continue to be supported) for at least three months of linear time from the rele
...
+than one release even if the minimum deprecation period has elapsed. For example,
+if a feature is deprecated in 0.20.0 which is released on January 20, 2022
+and the next minor version release 0.21.0 is released on June 16, 2022 the
+deprecated feature can't be remove until the 0.22.0 release even though
+0.21.0 was more than three months after the 0.20.0 release. This is important
+because the point of the deprecation warnings are to inform users that a
+potentially breaking API change is coming and to give them a chance to adapt
+their code. However, many users skip versions (especially if there are a large
+numbers of changes in each release) and don't upgrade to every release and
+might miss the warning if it's only present for a single minor version release.
```
jakelishman:
```suggestion
3a. A deprecated feature can not be removed unless it is deprecated in more
than one release even if the minimum deprecation period has elapsed. For example,
if a feature is deprecated in 0.20.0 which is released on January 20, 2022
and the next minor version release 0.21.0 is released on June 16, 2022 the
deprecated feature can't be removed until the 0.22.0 release, even though
0.21.0 was more than three months after the 0.20.0 release. This is important
because the point of the deprecation warnings are to inform users that a
potentially breaking API change is coming and to give them a chance to adapt
their code. However, many users skip versions (especially if there are a large
numbers of changes in each release) and don't upgrade to every release, so
might miss the warning if it's only present for a single minor version release.
```

docs/contributing_to_qiskit.rst
```diff
@@ -274,6 +274,15 @@ path. A migration path might be "stop using that feature", but in such cases
 it is necessary to first judge how widely used and/or important the feature
 is to users, in order to determine a reasonable obsolescence date.
 
+2a. The migration path must have existed in a least a prior release before the
+new feature can be deprecated. For example, if you have a function ``foo()``
+which is going to be replaced with ``bar()`` you can'd deprecate the ``foo()``
+function in the same release that introduces ``bar()``. The ``bar()`` function
```
kevinhartman, line 280: I'm not sure I'm completely following. Isn't the deprecation warning the thing that will inform most users that they ought to start thinking about a migration?

mtreinish: It is, but if you're maintaining a library built on qiskit if the alternative is introduced at the same time it means you're forced to raise the minimum dependency version to the latest release to avoid a deprecation warning.

This is actually something we hit in Qiskit in our consumption of matplotlib quite recently where they deprecated some APIs around 3d axes and the alternatives were introduced in the same version, which meant to fix the warnings we'd have to bump our minimum supported matplotlib version to that release which became quite a mess to maintain (especially as we hit bugs in how they did things in the alternatives).

Having this in the policy means we're giving people at least an `n - 1` (in `0.n.x` versions) slack in what they can expect to have working.

jakelishman:
Looks good to me, typos aside.  I don't know if we want to do it now, but it'd be good to have a section on `FutureWarning` for instances where a line of code will continue to be valid, but its result will change as well.

docs/contributing_to_qiskit.rst
mtreinish: Thinking about this a bit more it probably makes sense to point people to the `PendingDeprecationWarning` to use when the alternative is introduced to indicate that the API will be deprecated in the future. But that the actual deprecation timer starts a minimum of 1 release after the alternative is introduced.

jakelishman (approved):

PR #1378, opened 2021-12-10
Set IFS only locally for read in deploy_documentation_tags.sh

jakelishman:
### Summary

Previously IFS was set for the whole bash setting, which later affected
the word splitting when the secrets were read from the variables and
split into the openssl command.

This also tidies up the array reading - previously, the `VERSION`
variable was an array (so `echo $VERSION` would be equivalent to
`echo ${VERSION[0]}`), and it was passed to `cut`, which isn't
necessary, since it was already split.

### Details and comments

@mtreinish: I think this is the problem on the docs tag deploy action.

### DIFF
--- a/tools/deploy_documentation_tag.sh
+++ b/tools/deploy_documentation_tag.sh
@@ -25,8 +25,7 @@ pwd
 
 CURRENT_TAG=`git describe --abbrev=0`
 IFS=. read -ra VERSION <<< "$CURRENT_TAG"
-# The "minor" version, so in "0.29.2" it's "29".
-STABLE_VERSION=${VERSION[1]}
+STABLE_VERSION="${VERSION[0]}.${VERSION[1]}"
 
 # Build the documentation.
 tox -edocs -- -D content_prefix=documentation/stable/"$STABLE_VERSION" -j auto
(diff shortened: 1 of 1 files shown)


tools/deploy_documentation_tag.sh
```diff
@@ -24,9 +24,9 @@ echo "show current dir: "
 pwd
 
 CURRENT_TAG=`git describe --abbrev=0`
-IFS='.'
-read -ra VERSION <<< "$CURRENT_TAG"
-STABLE_VERSION=`echo $VERSION | cut -d "." -f -2`
+IFS=. read -ra VERSION <<< "$CURRENT_TAG"
+# The "minor" version, so in "0.29.2" it's "29".
+STABLE_VERSION=${VERSION[1]}
```
mtreinish: This needs to be:

```suggestion
STABLE_VERSION=${VERSION[0]}.${VERSION[1]}
```

I think because the publish url is `stable/0.33`. basically we want to publish stable versions at: qiskit.org/documentation/stable/major.minor

jakelishman: Thanks - I misread the `cut` command. I thought it was only picking out field `2`, as opposed to `-2`.  Fixed in 8534ac9cc.

jakelishman: And apparently didn't stop to think that that was kind of weird for a URL.

mtreinish (approved):

jakelishman:
I don't think we can manually trigger a re-run of the tags build with the new file until another release goes out, but that should go out with Aer 0.10 next week anyway.

PR #1384, opened 2021-12-15
Do not merge: Run tutorials on Python 3.9

jakelishman:
### Summary

There is a large performance regression in Aer 0.10, but locally it
seems to be limited to Python 3.7 and below.  This is testing if the
same is true on CI.

### Details and comments

Not for merging, just for testing.


jakelishman:
This PR did its job, and now with the release of Aer 0.10.1, tutorials should work anyway.

PR #1385, opened 2021-12-16
Bump Applications minimum versions

manoelmarques:
### Summary
Following the recent release 0.3.0 of all application modules, bumping their minimum versions in the meta-package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,22 +34,22 @@ requirements = [
 
 
 optimization_extra = [
-    "qiskit-optimization>=0.2.2",
+    "qiskit-optimization>=0.3.0",
 ]
 
 
 finance_extra = [
-    "qiskit-finance>=0.2.1",
+    "qiskit-finance>=0.3.0",
 ]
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
Looks fine to me.  Do we need a bump on Experiments to 0.2 as well?

manoelmarques:
@jakelishman I don't know why `qiskit-experiments` doesn't have a min.version test. We update the apps. in the meta-package in order to make it easier for the user to upgrade. I would think `qiskit-experiments` should follow the same principle. I can ammend this PR adding the test there as well.

jakelishman:
Sure - I'll bring it up with Chris to check, but chances are we'll want it, and I'll just add it in as part of the next release which will put out Aer 0.10 (and 0.10.1) and IBMQ-Provider (probably 0.18.3).

manoelmarques:
Thanks, just le me know in Slack if it is ok and I will add it to this PR.

PR #1401, opened 2022-01-21
Fix tutorial redirects being excluded from sync

jakelishman:
### Summary

* Correct the Sphinx configuration for the index files for the old
  tutorials directories; `sphinx-reredirect` wants the keys to be source
  files without extensions.

* Change the exclude patterns in the `rclone` script to be explicitly
  relative to the root directory only.  Previously, the tutorial
  redirects were being excluded from the sync (both for overwrite and
  delete) because `tutorials/optimization/index.html` is also excluded
  by the rule `- optimization/**`.  Instead, use the leading `/` to
  limit the rules to the root only.

### Details and comments

This should fix the old `optimization` tutorials still being present at https://qiskit.org/documentation/tutorials/optimization (and similar for finance, etc) - our Sphinx configuration is _meant_ to be redirecting them, but a slightly overzealous `rclone` exclude rule is preventing them from being synced.  You can see that the redirects are working in principle, because the stable branches (which aren't given the exclude filters) do correctly redirect, for example https://qiskit.org/documentation/stable/0.34/tutorials/optimization/.

Fix a part of Qiskit/qiskit-terra#10616.


mtreinish (approved):
Nice, I've known about the regex issue for a while (I found it when looking at https://qiskit.org/documentation/tutorials/machine_learning/02_qsvm_multiclass.html redirects but https://qiskit.org/documentation/tutorials/optimization) and forgot to fix it so thanks for doing this.

PR #1402, opened 2022-01-21
Expand platform support section to install doc

mtreinish:
### Summary

The supported platform list in the install documentation wasn't quite up
to date. The supported platforms are also a bit more nuanced than that
list made it appear. This commit expands the platform support section
and documents the different levels of support for platforms and the
details about each level. This should hopefully make it clearer to users
which platforms are expected to work and what guarantees are made for
each platform.

This is related to Qiskit/qiskit-terra#7553, Qiskit/qiskit-terra#7549,
and Qiskit/qiskit-aer#1362 which are expanding (or just changing the
level of support) the supported platforms but each have a different set
of tradeoffs that we should be documenting.


docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
 
       pip install -e .
 
+Platform Support
+----------------
+
+Qiskit strives to support as many platforms as possible but due to limitations
+in available testing resources and platform availability not all platforms
```
jakelishman:
```suggestion
Qiskit strives to support as many platforms as possible, but due to limitations
in available testing resources and platform availability, not all platforms
```

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+can be supported. Platform support for Qiskit is broken into 3 tiers with different
+levels of support for each tier. For platforms outside these Qiskit is probably
+still installable but it's not tested and you will have to build Qiskit (and likely
+Qiskit's dependencies) from source.
```
jakelishman:
```suggestion
levels of support for each tier. For platforms outside these, Qiskit is probably
still installable, but it's not tested and you will have to build Qiskit (and likely
Qiskit's dependencies) from source.
```

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+
+Additionally, Qiskit only supports CPython, running with other Python
```
jakelishman:
```suggestion
Additionally, Qiskit only supports CPython. Running with other Python
```

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+interpreters isn't currently supported.
+
+Tier 1
+''''''
+
+Tier 1 supported platforms are fully tested upstream as part of the development
+processes to ensure any proposed change will function correctly. Pre-compiled
+binaries are built, tested, and published to PyPI as part of the release process
```
jakelishman:
```suggestion
binaries are built, tested, and published to PyPI as part of the release process.
```

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+pre-compiled binaries are built, tested, and published to PyPI as part of the
+release process and these packages can be expected to be installed with just a
+functioning Python environment.
+
+Tier 2 platforms are currently:
+
+ * Linux i686 (distributions compatible with the
+   `manylinux 2010 <https://www.python.org/dev/peps/pep-0571/>`__ packaging
+   specification) for Python < 3.10
+ * Windows 32bit for Python < 3.10
```
jakelishman:
```suggestion
 * Windows 32 bit for Python < 3.10
```

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+ * Linux aarch64 (distributions compatible with the
+   `manylinux 2014 <https://www.python.org/dev/peps/pep-0599/>`__ packaging
+   specification)
+
+Tier 3
+''''''
+
+Tier 3 platforms are not tested upstream as part of the development process. However,
+pre-compiled binaries are built, but not tested, and published to PyPI as part of
+the release process. They may not be installable with just a functioning Python
```
jakelishman:
```suggestion
Tier 3 platforms are not tested upstream as part of the development process.  Pre-compiled
binaries are built and published to PyPI as part of the release process, with no
testing at all. They may not be installable with just a functioning Python
```

it's probably good to stress that we really don't test these at all, just build and pray

docs/getting_started.rst
```diff
@@ -557,6 +551,72 @@ Installation
...
+environment and may require a C/C++ compiler or additional programs to build
```
mtreinish, line 608: I probably should mention a fortran compiler here too, because building scipy from source might need a fortran compiler for lapack.

jakelishman (commented):
This looks like a really good addition to the documentation, thanks!  The text is already pretty clear to me, I just left a couple of super nitpicky bits.

docs/getting_started.rst
jakelishman: I thought about that, but then you might need to say a rust compiler too for retworkx, and it all just starts getting a bit much.  I maybe wouldn't worry too much.

mtreinish: Well retworkx actually has tier 1 or tier 2 support for all these platforms and ppc64le and s390x (except for macOS arm64 which is still tier 3 because there's no way to run that in CI). So you shouldn't need a rust compiler for anything in this list. But point taken this is probably sufficient.

docs/getting_started.rst
mtreinish: Heh, yeah a good idea.

jakelishman (approved):

PR #1405, opened 2022-01-25
Add seed to all transpiler benchmarks

### LINKED ISSUE #1404: RippleAdder benchmarks don't set `seed_transpiler`

The benchmarks that live in `test/benchmarks/ripple_adder.py` don't set a value for `seed_transpiler`, resulting in noisy outputs that can hide small but significant changes. e.g. in `track_depth_transpile_square_grid_ripple_adder`, only https://github.com/Qiskit/qiskit-terra/pull/7205 is visible. https://qiskit.github.io/qiskit/#ripple_adder.RippleAdderTranspile.track_depth_transpile_square_grid_ripple_adder?commits=2fe852fa

jakelishman:
### Summary

Fill in a missing seed for all benchmarks that need one.

### Details and comments

Fix #1404.

### DIFF
--- a/test/benchmarks/transpiler_benchmarks.py
+++ b/test/benchmarks/transpiler_benchmarks.py
@@ -61,17 +61,20 @@ class TranspilerBenchSuite:
     def time_single_gate_compile(self):
         circ = qiskit.compiler.transpile(self.single_gate_circuit,
                                          coupling_map=self.coupling_map,
-                                         basis_gates=self.basis)
+                                         basis_gates=self.basis,
+                                         seed_transpiler=20220125)
         qiskit.compiler.assemble(circ)
 
     def time_cx_compile(self):
         circ = qiskit.compiler.transpile(self.cx_circuit,
                                          coupling_map=self.coupling_map,
-                                         basis_gates=self.basis)
+                                         basis_gates=self.basis,
+                                         seed_transpiler=20220125)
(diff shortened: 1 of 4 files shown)


test/benchmarks/transpiler_benchmarks.py
```diff
@@ -61,17 +61,20 @@ def setup(self):
     def time_single_gate_compile(self):
         circ = qiskit.compiler.transpile(self.single_gate_circuit,
                                          coupling_map=self.coupling_map,
-                                         basis_gates=self.basis)
+                                         basis_gates=self.basis,
+                                         seed_transpiler=20220125)
```
mtreinish, line 65: I'd be really surprised if this has an impact on the output for this case since it's a single h gate on 1 qubit so there is no routing and initial layout is always used, But it won't hurt at all either

mtreinish (approved):
This LGTM, it will invalidate some results (unless the benchmark has an explicit version set). But I think that's ok since we weren't finding meaningful regressions before anyway so the loss of historical data isn't important.

test/benchmarks/transpiler_benchmarks.py
jakelishman: Yeah, I just searched for all instances of `transpile` and stuck a seed on them - it's probably better to put it in place now, just in case some future change injects randomness in this case.

PR #1409, opened 2022-02-01
grammatical mistakes

DRA-chaos:
Summary
Grammatical errors

Details and comments
Rectified a few errors related to sentence structures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,11 @@
 
 **Qiskit** is an open-source SDK for working with quantum computers at the level of circuits, algorithms, and application modules.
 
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
Thanks for the interest in contributing!

Really, none of these were actually incorrect English. For the first one, "at one place" is definitely incorrect and says something different - the point is that you need only one command, not that they all get put in the same place.

The other places are both correct English already ("of installing" and "to install" are both fine), and in the second, "optional extra" is the standard terminology for this.

I'll close this, but please do feel free to look through the "help wanted" issues on all of the Qiskit family repositories if you're interested in contributing.

PR #1421, opened 2022-02-16
Add benchmarks for circuit scheduling

itoko:
### Summary
Add benchmarks to track the time performance of circuit scheduling

### Details and comments
Add four tests, which benchmark TimeUnitConversion, ALAPSchedule, ASAPSchedule and DynamicalDecoupling passes. It would be a good time to start benchmarking those as rework on circuit scheduling passes is about to start.

### DIFF
--- a/test/benchmarks/scheduling_passes.py
+++ b/test/benchmarks/scheduling_passes.py
@@ -13,14 +13,18 @@
 # that they have been altered from the originals.
 
 # pylint: disable=invalid-name,missing-docstring
-# pylint: disable=attribute-defined-outside-init,
-# pylint: disable=unused-wildcard-import,wildcard-import
+# pylint: disable=attribute-defined-outside-init
 
 from qiskit import transpile
 from qiskit.circuit.library.standard_gates import XGate
 from qiskit.transpiler import CouplingMap
 from qiskit.transpiler import InstructionDurations
-from qiskit.transpiler.passes import *
+from qiskit.transpiler.passes import (
+    TimeUnitConversion,
+    ASAPSchedule,
+    ALAPSchedule,
+    DynamicalDecoupling,
+)
 from qiskit.converters import circuit_to_dag
 
 from .utils import random_circuit
--- a/test/benchmarks/scheduling_passes.py
+++ b/test/benchmarks/scheduling_passes.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=invalid-name,missing-docstring
+# pylint: disable=attribute-defined-outside-init,
+# pylint: disable=unused-wildcard-import,wildcard-import
+
+from qiskit import transpile
+from qiskit.circuit.library.standard_gates import XGate
+from qiskit.transpiler import CouplingMap
+from qiskit.transpiler import InstructionDurations
+from qiskit.transpiler.passes import *
+from qiskit.converters import circuit_to_dag
+
+from .utils import random_circuit
+
+
+class SchedulingPassBenchmarks:
+
+    params = ([5, 10, 20],
+              [500, 1000])
+    param_names = ['n_qubits', 'depth']
+    timeout = 300
+
+    def setup(self, n_qubits, depth):
+        seed = 42
+        self.circuit = random_circuit(n_qubits, depth, measure=True,
+                                      conditional=True, reset=True, seed=seed,
+                                      max_operands=2)
+        self.basis_gates = ['rz', 'sx', 'x', 'cx', 'id', 'reset']
+        self.cmap = [[0, 1], [1, 0], [1, 2], [1, 6], [2, 1], [2, 3], [3, 2],
+                     [3, 4], [3, 8], [4, 3], [5, 6], [5, 10], [6, 1], [6, 5],
+                     [6, 7], [7, 6], [7, 8], [7, 12], [8, 3], [8, 7], [8, 9],
+                     [9, 8], [9, 14], [10, 5], [10, 11], [11, 10], [11, 12],
+                     [11, 16], [12, 7], [12, 11], [12, 13], [13, 12], [13, 14],
+                     [13, 18], [14, 9], [14, 13], [15, 16], [16, 11], [16, 15],
+                     [16, 17], [17, 16], [17, 18], [18, 13], [18, 17],
+                     [18, 19], [19, 18]]
+        self.coupling_map = CouplingMap(self.cmap)
+        self.transpiled_circuit = transpile(self.circuit,
+                                            basis_gates=self.basis_gates,
+                                            coupling_map=self.coupling_map,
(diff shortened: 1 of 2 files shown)


itoko:
Here is an example of the results of new benchmarks in my local env.
```
[ 66.67%] ··· ...Benchmarks.time_alap_schedule_pass                 ok
[ 66.67%] ··· ========== =========== ============
              --                  depth          
              ---------- ------------------------
               n_qubits      500         1000    
              ========== =========== ============
                  5       102±0.6ms    202±3ms   
                  10       278±7ms     605±10ms  
                  20       844±40ms   1.72±0.02s 
              ========== =========== ============

[ 83.33%] ··· ...Benchmarks.time_asap_schedule_pass                 ok
[ 83.33%] ··· ========== ========== ============
              --                  depth         
              ---------- -----------------------
               n_qubits     500         1000    
              ========== ========== ============
                  5       110±5ms     201±2ms   
                  10      306±2ms     604±7ms   
                  20      901±10ms   1.85±0.01s 
              ========== ========== ============

[100.00%] ··· ...rks.time_dynamical_decoupling_pass                 ok
[100.00%] ··· ========== ============ ============
              --                   depth          
              ---------- -------------------------
               n_qubits      500          1000    
              ========== ============ ============
                  5        118±3ms      233±7ms   
                  10       339±7ms      684±5ms   
                  20      1.06±0.01s   2.09±0.05s 
              ========== ============ ============
```

itoko:
```
[100.00%] ··· ...rks.time_time_unit_conversion_pass                 ok
[100.00%] ··· ========== ========== ==========
              --                 depth        
              ---------- ---------------------
               n_qubits     500        1000   
              ========== ========== ==========
                  5       51.3±2ms   110±6ms  
                  10      141±2ms    296±2ms  
                  20      414±3ms    837±10ms 
              ========== ========== ==========
```
```
[100.00%] ··· ...enchmarks.time_schedule_qv_14_x_14                 ok
[100.00%] ··· =============================== ============
               transpiler optimization level              
              ------------------------------- ------------
                             0                  628±10ms  
                             1                 1.40±0.03s 
                             2                 1.86±0.02s 
                             3                 3.63±0.05s 
              =============================== ============
```

test/benchmarks/scheduling_passes.py
```diff
@@ -0,0 +1,87 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=invalid-name,missing-docstring
+# pylint: disable=attribute-defined-outside-init,
+# pylint: disable=unused-wildcard-import,wildcard-import
```
jakelishman: There is a reason not to do wildcard imports - it's more difficult for IDEs to reason about code, and it's harder to read code if you can't immediately tell where an import came from. It'd be better to explicitly import the names you need from `transpiler.passes`, especially since there's so few of them.

test/benchmarks/transpiler_levels.py
```diff
@@ -197,3 +206,9 @@ def time_transpile_qv_14_x_14(self, transpiler_level):
     def track_depth_transpile_qv_14_x_14(self, transpiler_level):
         return transpile(self.qv_14_x_14, self.melbourne, seed_transpiler=0,
                          optimization_level=transpiler_level).depth()
+
+    def time_schedule_qv_14_x_14(self, transpiler_level):
+        transpile(self.qv_14_x_14, self.melbourne, seed_transpiler=0,
+                  optimization_level=transpiler_level,
+                  scheduling_method="alap",
+                  instruction_durations=self.durations)
```
jakelishman: Given that we're already testing the scheduling passes separately, is there a need to add this on as well?  These are expensive benchmarks to run, and a good portion of the time is just going to be spent doing the same benchmarking as `time_transpile_qv_14_x_14`.

itoko: OK, I liked to have a bench to measure the time near to that end users actually experience when scheduling, but I see the cost is going to be greater than the benefit. I'll remove this.

itoko:
@jakelishman I've updated according to your suggestions. Would you mind taking a second look?

test/benchmarks/transpiler_levels.py
itoko: Done in [883d335](https://github.com/Qiskit/qiskit/pull/1421/commits/883d335f7211cd2101bf152e68a937f40eaf9b45)

jakelishman (approved):
Thanks for the update, and sorry for the delay in second review.

This looks fine from my perspective, but I don't know enough about the targets for the benchmark suite to give final approval.  I'll ask @kdk (who's currently away) or @mtreinish to check, and particularly check if the suggestion I made in https://github.com/Qiskit/qiskit/pull/1421#discussion_r807713770 was appropriate - I should have made it clearer that that was a question rather than a concrete suggestion!

itoko:
Sorry to cut in, what about an idea in the middle? That is introducing `measure_pauli` accepting multi-qubit pauli measurement while it internally produces single-qubit `PauliMeasure`s for now to avoid the drawer issue for mult-clbits. The direct creation of multi-qubit pauli measurement, e.g. `PauliMeasure(basis="XYZ")`, should raise "NotImplementedError" for now. In this way, once the drawer issue is fixed, we can introduce true multi-qubit pauli measurement in the future without a big change. Am I missing something?

test/benchmarks/scheduling_passes.py
```diff
@@ -0,0 +1,91 @@
...
+                                      max_operands=2)
+        self.basis_gates = ['rz', 'sx', 'x', 'cx', 'id', 'reset']
+        self.cmap = [[0, 1], [1, 0], [1, 2], [1, 6], [2, 1], [2, 3], [3, 2],
+                     [3, 4], [3, 8], [4, 3], [5, 6], [5, 10], [6, 1], [6, 5],
+                     [6, 7], [7, 6], [7, 8], [7, 12], [8, 3], [8, 7], [8, 9],
+                     [9, 8], [9, 14], [10, 5], [10, 11], [11, 10], [11, 12],
+                     [11, 16], [12, 7], [12, 11], [12, 13], [13, 12], [13, 14],
+                     [13, 18], [14, 9], [14, 13], [15, 16], [16, 11], [16, 15],
+                     [16, 17], [17, 16], [17, 18], [18, 13], [18, 17],
+                     [18, 19], [19, 18]]
```
mtreinish, line 53: I'm curious which coupling map is this? Too many edges for me to visualize it in my head.

mtreinish (approved):
This LGTM, I think having a full path scheduling benchmark makes sense just to see what the overhead is in an end-to-end benchmark. The synthetic per pass benchmarks only take you so far. The 14x14 qv is a good choice for that because it's relatively quick. But @jakelishman is correct that we do have a time budget to worry about too. I haven't checked what our current time per commit is on the nightly runs I think we'd be ok adding a couple of optimization levels to the suite. If we added like optimization level 0 and 1 (as I think those are the likely place where people are doing scheduling) that would only add maybe 30sec-1min per commit (as each benchmark does at least 4 runs sometimes more). But I think we can do this in a follow up PR as it's easy enough to add more benchmarks

test/benchmarks/scheduling_passes.py
itoko: To be honest, I just copied it from `PassBenchmarks`(`mapping_passes`) to make the comparison with other benchmarks some sense, and I don't know where it comes from.

mtreinish: Heh, well I wrote those benchmarks but a long time ago. I think I would have used either the Tokyo or Singapore coupling maps there but it's just a vague memory.

PR #1437, opened 2022-03-01
Update asv.conf.json to support recent changes

mtreinish:
### Summary

This commit updates the asv configuration to support two recent changes
in the terra repo. The first is updating the supported python version
list to reflect the current versions supported by terra. Python 3.6 is
no longer supported and python 3.10 is now supported. Additionally,
after Qiskit/qiskit-terra#7658 merged setuptools-rust and Rust are now
being used to build compiled extensions. While cython is still being
used, it's use will be removed soon with Qiskit/qiskit-terra#7702. This
commit updates the build configuration to build the rust extension and
then build a wheel from it instead of building the cython extension.

### Details and comments

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -35,8 +35,8 @@
         "return-code=any python -mpip uninstall -y qiskit-ignis"
     ],
     "build_command": [
-        "pip install -U Cython",
-        "python setup.py build",
+        "pip install -U Cython setuptools-rust",
+        "python setup.py build_rust --release",
         "PIP_NO_BUILD_ISOLATION=false python -mpip wheel --no-deps --no-index -w {build_cache_dir} {build_dir}"
     ],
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):

jakelishman:
Semi-unrelated, but is there a path through the build that does what we need, but respects the `pyproject.toml` `build-requires` section?  Feels like we oughtn't to need to manually specify Cython and `setuptools_rust`.

mtreinish:
If we drop the explicit `python setup.py build*` call we can respect the `pyproject.toml` and leverage pep 518 to install setuptools rust. For this I mostly was just adapting what we already had which predates our use of a pyproject.toml in terra I can update it if you'd prefer to just modernize the whole thing now

jakelishman:
No, I'm happy to do that later - I only commented on it now because I was looking at it.

PR #1441, opened 2022-03-04
Add overall transpilation time bench with scheduling

itoko:
### Summary
Add a benchmark to track the overall time performance of circuit scheduling, which would be near to that end users actually experience when scheduling.

### Details and comments
Follow-up PR of #1421. As suggested in the original PR, optimization level 0 and 1 are added so that it would not take so much time while covering the places where people are doing scheduling.

### DIFF
--- a/test/benchmarks/transpiler_levels.py
+++ b/test/benchmarks/transpiler_levels.py
@@ -19,6 +19,7 @@ import os
 
 from qiskit.compiler import transpile
 from qiskit import QuantumCircuit
+from qiskit.transpiler import InstructionDurations
 
 from .backends.fake_melbourne import FakeMelbourne
 from .utils import build_qv_model_circuit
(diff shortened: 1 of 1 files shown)


test/benchmarks/transpiler_levels.py
```diff
@@ -197,3 +206,10 @@ def time_transpile_qv_14_x_14(self, transpiler_level):
     def track_depth_transpile_qv_14_x_14(self, transpiler_level):
         return transpile(self.qv_14_x_14, self.melbourne, seed_transpiler=0,
                          optimization_level=transpiler_level).depth()
+
+    def time_schedule_qv_14_x_14(self, transpiler_level):
+        if transpiler_level <= 1:
```
itoko: Any better way to skip `transpiler_level >= 2` cases?

mtreinish: The easiest way is after the function definition you can do something like: `time_schedule_qv_14_x_14.params = [0, 1]` which will override the params list for this method.

mtreinish (approved):

PR #1455, opened 2022-03-31
fix url error

NG-Glen:
apidoc/aer_provider.html


HuangJunye (approved):
Good catch! Thanks for submitting the PR to fix.

jakelishman (approved):

PR #1457, opened 2022-04-01
adopt a accurate wording

NG-Glen:


jakelishman:
We are overhauling this entire guide elsewhere, but in this case the word "issue" is used because it's tab you need to click on in GitHub, which makes it clearer what's meant.  Thanks for your interest, and please look at the "good first issues" tags on the different repositories in this organisation for some places where we're particularly interested in receiving contributions.

jakelishman:
We are overhauling this entire guide elsewhere, but in this case the word "issue" is used because it's tab you need to click on in GitHub, which makes it clearer what's meant.  Thanks for your interest, and please look at the "good first issues" tags on the different repositories in this organisation for some places where we're particularly interested in receiving contributions.

PR #1463, opened 2022-04-08
ini file to ini format file

NG-Glen:


HuangJunye:
@NG-Glen Thanks for your contribution. However, I am not sure such a change is required.

NG-Glen:
> @NG-Glen Thanks for your contribution. However, I am not sure such a change is required.

or I think a more accurate description is property file, windows user will find the .ini extension name file, but is who is developer who will more understand property file.

PR #1361, opened 2021-11-19
Remove qiskit-aqua from the metapackage

mtreinish:
### Summary

The next qiskit-metapackage release will no longer include qiskit-aqua
as the deprecation cycle has finished. This commit removes the
qiskit-aqua package from the metapackage, deletes the documentation
sections for aqua and adds a prelude section for the release notes
documenting the removal and manual upgrade steps users might need to do.

### Details and comments

PR #1466, opened 2022-04-11
removing Ignis from contributing_to_qiskit

1ucian0:
The [Contributing to Qiskit page](https://qiskit.org/documentation/contributing_to_qiskit.html) still has some references to Ignis.


mtreinish (commented):
Ignis is still part of the metapackage while deprecated. I've been waiting to do this until we actually remove it (which will be after the next experiments release). You can see when we did this for aqua: https://github.com/Qiskit/qiskit/pull/1361

PR #1467, opened 2022-04-12
conda list package is linked not active

NG-Glen:
Please reference conda help


jakelishman:
The existing text here is fine, even if not precisely what Anaconda themselves use.  I've left more information about how to find potential contributions in comments on some of your other PRs, and thanks for your interest.

PR #1468, opened 2022-04-12
Jupyter Notebook (formerly IPython Notebooks)

NG-Glen:


garrison:
The name Jupyter has been in use for eight years now.  Why do you suggest also including the old name?

jakelishman:
While we always appreciate honest contributions, this is one of many single-word documentation PRs you've made to various repos across the Qiskit organisation that are questionable at best. If you would like to contribute, please take a look at issues tagged "good first issue" on the various repositories, and please could you stop with the changes to text that does not need changing.

PR #1471, opened 2022-04-13
Write separate deprecation guide

jakelishman:
### Summary

This moves the deprecation policy into a separate file, and rewrites the
text to be shorter and more split it.  Also add sections on the
difficulty of changing behaviours and how to potentially avoid those
situations, and how to test the deprecated functionality correctly.

The actual policy should be unchanged by this commit.


jakelishman:
Whoops, this was meant to be against my own fork.

PR #1487, opened 2022-05-04
Remove qiskit-ignis from the metapackage

### LINKED ISSUE #1485: Migriate ignis-based benchmarks to qiskit-experiments

The Qiskit benchmarks in `test/benchmarks` have several benchmarks which rely on code from Ignis to build, compile, and simulate circuits, and analyze results (e.g https://github.com/Qiskit/qiskit/blob/9d2359525d34e7ab67b1ac00dc5aa3cfedf4f2ff/test/benchmarks/state_tomography.py#L24 and https://github.com/Qiskit/qiskit/blob/684dfe5dbf25caff27081fe3d78331837d908f72/test/benchmarks/randomized_benchmarking.py#L63 ). With the release of qiskit-experiments, these should be updated to use the corresponding methods, and likely expanded to cover some use-cases from qiskit-experiments.

@gadial and @chriseclectic Any thoughts on what would pieces of experiments would be useful to benchmark here?

mtreinish:
### Summary

The next qiskit metapackage release (when terra releases 0.21.0) will
no longer include qiskit-ignis as the deprecation cycle has finished.
This commit removes the qiskit-ignis package from the metapackage,
deletes the documentation sections for ignis and adds a prelude
section for the release notes documenting the removal and manual
upgrade steps users might need to do.

### Details and comments

We should not merge this until we're ready to release the next minor version of the metapackage (which will correspond with the qiskit-terra 0.21.0 release). This also gives us time to resolve #1485 and also update the migration guide to be more thorough before we drop this.


jakelishman (commented):
Mostly looks fine.  Do we need to pull the link in the toctree of `docs/index.rst` as well?

I don't know what we did for Aqua, but is it worth putting some note on the main documentation page somewhere (not just release notes) saying "Ignis is deprecated, here's the migration guide"?

HuangJunye:
I think a migration guide for Ignis will be great!

jakelishman:
@HuangJunye: there already is [a migration guide for Ignis](https://github.com/Qiskit/qiskit-ignis#migration-guide) - it's linked in the release note added in this PR, I was more asking about whether we should make that more visible.

HuangJunye:
I see. I think it would be good to make it more obvious for a while at least (until next release?). I think we had it as one of the side bar item for Aqua migration guide. But Ignis has a lot fewer users than Aqua so I am not sure that is needed.

HuangJunye (approved):
LGTM!

mtreinish:
We're going to release qiskit 0.37.0 today so I'm removing the on hold here since we should be ready to merge this now to remove ignis as part of the release.

PR #1491, opened 2022-05-18
Windows 32 bit and Windows 32bit, format to same spell

NG-Glen:
Windows 32 bit and Windows 32bit, format to same spell

PR #1498, opened 2022-05-18
Add CODEOWNERS and Mergify configuration

jakelishman:
### Summary

This will let us restrict direct write access to the metapackage, while
empowering more people to approve and merge pull requests that they are
reponsible for maintaining.  Provided the mergify bot has the required
write access to create the merge commits, we can give non-administrator
code owners the "Triage" permission only, in order to add the
`automerge` label, and all commit merges will go through the Mergify
bot.

### Details and comments

I wasn't sure who to add in the various sections here, so open for more discussion.

The PR seems to be saying that Junye is insufficiently privileged (because he doesn't currently have write access), even though [the documentation says individually code owners only need to have read permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners):

> The people you choose as code owners must have read permissions for the repository.

This text changed in February of this year (probably about the same time the linter was introduced), but the linter on the file below seems to be complaining anyway.  I don't know which will actually take effect if we were to merge it.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,13 @@
+# This file defines the set of people who are responsible for different parts of
+# the Qiskit metapackage code.
+#
+# Those assigned as owners of the whole repository are the core maintainers, who
+# (using privileges granted by other means) also have the power to make releases
+# of the metapackage.  These people can approve all pull requests.
+* @mtreinish @jakelishman @kevinhartman
+
+# More specialised rules override the global rule, so the core team need to be
+# respecified.
+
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
Let's see which is correct the linter or the docs. It's easy to update the permission level for codeowners if this doesn't work with the current permission levels

mtreinish:
I'm testing this on #1491 and Junye isn't recognized as a codeowner/Approved reviewer at his current permission level. So the linter is correct in what it was checking and the docs are wrong. We'll have to update him to have write permissions so the codeowner holds

PR #1494, opened 2022-05-18
Remove ignis tutorials section

HuangJunye:
### Summary

Remove ignis tutorials section as part of the deprecation cycle. 

### Details and comments
Fixes #1492

PR #1500, opened 2022-05-18
Add missing redirects for ignis tutorials

mtreinish:
### Summary

In #1494 the index page for the removed ignis tutorials were removed,
this commit also needed to add redirects so that any saved links don't
404. This commit adds the missing redirects so that the tutorials are
redirected to approriate new package's documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -108,6 +108,25 @@ ml_tutorials = [
     'index'
 ]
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
All the defined redirects look right to me - they're the same names as those removed in #1309 and the redirects look fine.

PR #1502, opened 2022-05-27
Bump applications minimum versions

manoelmarques:
### Summary
Bump the applications minimum versions to the current versions in Pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ optimization_extra = [
 
 
 finance_extra = [
-    "qiskit-finance>=0.3.1",
+    "qiskit-finance>=0.3.2",
 ]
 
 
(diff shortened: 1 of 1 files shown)


jakelishman (dismissed):
This probably ought to go out as Qiskit 0.37.0 (in early July), coinciding with the release of Terra 0.21, since it contains a minor-version bump to machine-learning. (Not that the extras need to dictate the versioning entirely.) Alternatively, if Aer 0.11 is ready before then, it can go out with that.

I'll approve but leave unmerged til then, just in case something unforseen happens and we need to release a patch version before another minor.

jakelishman (approved):
Is this the last likely change to the application versions?  If so, let's merge now, because the Terra release is on schedule for tomorrow.

manoelmarques:
> Is this the last likely change to the application versions? If so, let's merge now, because the Terra release is on schedule for tomorrow.

Yes, there won’t be any more releases for now. Thanks

PR #1565, opened 2022-07-13
pep-0599 superseded by pep-0600

jwoehr:
Issue reported in [Qiskit Slack](https://qiskit.slack.com/archives/CSUDA6KM1/p1657705406991269)

### Summary
pep-0599 referenced in manylinux 2014 link superseded by pep-0600
Issue reported in [Qiskit Slack](https://qiskit.slack.com/archives/CSUDA6KM1/p1657705406991269)

### Details and comments
Updated doc to ref pep-0600
Trivial, did not put in any relnote for this

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -536,7 +536,7 @@ environment as all dependencies are available on these platforms.
 Tier 1 platforms are currently:
 
  * Linux x86_64 (distributions compatible with the
-   `manylinux 2014 <https://www.python.org/dev/peps/pep-0599/>`__
+   `manylinux 2014 <https://peps.python.org/pep-0600/>`__
    packaging specification.
  * macOS x86_64 (10.9 or newer)
  * Windows 64 bit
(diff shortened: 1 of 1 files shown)


mtreinish (changes requested):
This actually isn't correct. We're building packages to the manylinux 2014 spec which outlines a specific environment for the tag (basically just centos 7). pep600 is a more general future facing specification that explains how to do tags based on the platform's libc version, etc instead of being pinned to a specific environment. For the documentation linking to pep 599 is what we want to link people to because it outlines exactly what environment and compatibility level we're building binaries to.

I think the specific details from the slack conversation you linked in the PR summary is actually that the installation section is blank on the hosted docs: https://qiskit.org/documentation/getting_started.html#installation and the details are missing on how to install qiskit. I'm assuming there is a sphinx warning caused by a syntax error in this file that's preventing the instllation docs from being included in the compiled output.

jwoehr:
Ah, thanks. I'll close this.

PR #1567, opened 2022-07-14
Update tab syntax for sphinx-design

### LINKED ISSUE #1566: "Installation" section of "Getting started" page is empty
### Informations

- **Qiskit version**: NA
- **Python version**: NA
- **Operating system**: NA

[Getting started](https://qiskit.org/documentation/getting_started.html) page has "Installation" section. But, it looks empty.
image

The source rst file contains some texts.
https://github.com/Qiskit/qiskit/blob/master/docs/getting_started.rst#installation

access: https://qiskit.org/documentation/getting_started.html

Show tabs in "Installation" section of https://qiskit.org/documentation/getting_started.html

### Suggested solutions

No idea.

mtreinish:
### Summary

In #1558 the sphinx plugin sphinx-panel was switched to sphinx-design to
enable building the documentation with sphinx >=5.0.0. However, while
the build succeeded there were warnings emitted about tabbed directive
which was only valid for sphinx-panel. This prevented any tabbed
sections in the documentation from being built as sphinx just ignored
it (this is why it's best practice to build with -W to make warnings
fatal because they're almost always pointing to broken docs). This
commit fixes this by migrating the directives to use the syntax from
sphinx-design so that we're building the tabs sections of the
documentation.

### Details and comments

Fixes #1566

### DIFF
--- a/docs/contributing_to_qiskit.rst
+++ b/docs/contributing_to_qiskit.rst
@@ -696,43 +696,45 @@ from system-wide packages. This way, we avoid inadvertently becoming dependent o
 particular system configuration. For developers, this also makes it easy to maintain multiple
 environments (e.g. one per supported Python version, for older versions of Qiskit, etc.).
 
-.. tabbed:: Python venv
+.. tab-set::
 
-   All Python versions supported by Qiskit include built-in virtual environment module
-   `venv <https://docs.python.org/3/tutorial/venv.html>`__.
+    .. tab-item:: Python venv
 
-   Start by creating a new virtual environment with ``venv``. The resulting
-   environment will use the same version of Python that created it and will not inherit installed
-   system-wide packages by default. The specified folder will be created and is used to hold the environment's
-   installation. It can be placed anywhere. For more detail, see the official Python documentation,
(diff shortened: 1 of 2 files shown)


HuangJunye (commented):
Thanks for the fix. It looks good to me. However it my local build I have issues building it without `sphinx-panel`. It looks like `qiskit-metal` still uses `sphinx-panel`. However I am not familiar with the meta package documentation building process. I am not sure if fixing `qiskit-metal` is in the scope of this PR.

EDIT. Found the same warning for qiskit-metal on CI

https://github.com/Qiskit/qiskit/runs/7342031456?check_suite_focus=true#step:6:5572

mtreinish:
It's not in scope for this PR. This is only about the metapackage documentation which is anything included in the metapackage (and shares a common root `qiskit.*` namespace). Right now that's terra, aer, and the ibmq provider. All other projects have independent documentation and self contained builds. An issue in qiskit-metal will need to be fixed separately in their repo (and they should have `-W` set so those warnings are caught, we can't do that here in the metapackage because of integration issues between the 3 repositories).

HuangJunye (approved):

PR #1579, opened 2022-07-28
Separate stable version list from language menu

### LINKED ISSUE #1266: Move Previous Releases to a more visible location
**Source file:**
https://qiskit.org/documentation/

**Thank You**
Thanks to [#1261](https://github.com/Qiskit/qiskit/pull/1261) we now have previous releases available, which is great. 

**Issue:**
Currently, the Previous Releases (0.24.0, 0.25.0, 0.26.0) are only accessible from the language switcher drop-down and not easy to find. 

**Proposed solution:**
Show the Previous Releases in the left hand pane where it is visible. For example, under the Migration Guide or Trouble Shooting.
image

mtreinish:
### Summary

The language menu is using a widget/theme element rst-versions which in
the original rtd theme (which is the grandparent of the qiskit sphinx
theme) is a pop up/dropdown menu that has a bunch of extra information
in it that can be used to list different versions of the documentation
as well as download links. In #1261 that functionality was used to add a
list of stable versions to the sidebar since it was the logical place
for it. However, this has apparently been a source of confusion for most
users as unless they have prior knowledge about the previous version
links being on the same menu as the language selection and the
availability of old versions of the documentation goes unnoticed. This
commit addresses this by splitting the previous versions list out of the
language selector drop down and just adding a list to the bottom of the
side bar. We can change the formatting and style of this as needed in
the future, but at least for right now it gives us a clear location for
links to old version of the documenation.

### Details and comments
Fixes #1266

### DIFF
--- a/docs/_templates/stable_versions.html
+++ b/docs/_templates/stable_versions.html
@@ -1,7 +1,7 @@
 <div>
     <dl>
       <dt>Previous Releases</dt>
-      {% for version in version_list %}
+      {% for version in version_list | sort(reverse=True) %}
         <dd><a class="version" href="/documentation/stable/{{ version }}/index.html">{{ version }}</a></dd>
       {% endfor %}
     </dl>
--- a/docs/_templates/stable_versions.html
+++ b/docs/_templates/stable_versions.html
@@ -1,0 +1,8 @@
+<div>
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
+        <dd><a class="version" href="/documentation/stable/{{ version }}/index.html">{{ version }}</a></dd>
+      {% endfor %}
+    </dl>
+</div>
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -168,7 +168,6 @@
   {% block extrabody %} {% endblock %}
 
     {# SIDE NAV, TOGGLES ON MOBILE #}
-
     {% include "versions.html" %}
 
     <div class="table-of-contents-link-wrapper">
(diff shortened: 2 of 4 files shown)


mtreinish:
This is what the side bar looks like when I build this locally:

Screenshot_2022-07-28_13-34-09

and with the language menu selected

Screenshot_2022-07-28_13-34-33

docs/_templates/stable_versions.html
```diff
@@ -0,0 +1,8 @@
+<div>
+    <dl>
+      <dt>Previous Releases</dt>
+      {% for version in version_list %}
```
1ucian0: I think it would make sense to have newer first:
```suggestion
      {% for version in version_list | sort(reverse=True) %}
```

Probably, the right way of thing this is in `_get_version_list` with `StrictVersion` because semantic versioning:
```
from distutils.version import StrictVersion
...
    version_list.sort(key=StrictVersion)
    return version_list
```

1ucian0 (changes requested):
just a small comment. Thanks for taking care of this!

docs/_templates/stable_versions.html
mtreinish: I think we should be ok because even with semver I can't think of a case where this will sort incorrectly just using a string sort of the version number. The only version tag I could see having a potential issue is if we did prereleases where `0.2.0` would be less than `0.2.0rc0`. But for the metapackage we don't do pre-releases this won't come up.

But I think pushing a follow up to use [packaging](https://packaging.pypa.io/en/latest/index.html)  (which replaced the deprecated `distutils.version` module) is a good idea.

1ucian0 (approved):

Issue #1508, opened 2022-06-02
Aer needs its own documentation page


1ucian0:
With the [removal of Aer from the metapackage](https://github.com/Qiskit/qiskit-aer/pull/1526), a documentation page for its own is needed.

Once Aer is removed from the metapackage, the module will be at the same level of other application modules and it will require its own https://qiskit.org/documentation/aer/ in the same way that https://qiskit.org/documentation/nature/ exists, for example.

I suggest creating https://qiskit.org/documentation/aer/ silently with no change in the metapackage documentation first.

mtreinish:
Just for clarification that aer pr does not remove aer from the metapackage. It is creating a new python namespace. The term package there is in the strict python terminology which is confusing for most people: https://docs.python.org/3/glossary.html#term-package

As for the issue this is kinda premature. Aer is still part of the metapackage for now. This issue is also primarily not something we fix in the metapackage repository as publishing self contained docs is done in the aer repository.

mtreinish:
I apologize "premature" probably wasn't the correct way to describe this. Upon reflection renaming the qiskit-aer package from `qiskit.providers.aer` -> `qiskit_aer` is going to have some complications for link preservation. Since when we merge Qiskit/qiskit-aer#1526 the aer docs no longer refer to the old location. We'll probably need to use `sphinx-reredirects` to redirect from the old path to the new one (which likely means indexing everything and redirecting manually, see: https://github.com/Qiskit/retworkx/blob/main/docs/source/conf.py#L125-L128 and https://github.com/Qiskit/retworkx/blob/main/docs/source/sources.txt for what I did in retworkx to go from `retworkx/stubs/*` -> `retworkx/apiref/*`). I think having aer publish a standalone documentation page for it's next release sounds like the best course of action. That's primarily an aer issue, but doing the cleanup for the metapackage side of the docs we can track here.

mtreinish:
I've pushed up https://github.com/Qiskit/qiskit-aer/pull/1589, https://github.com/Qiskit/qiskit/pull/1595, and https://github.com/Qiskit/qiskit/pull/1596 which should address this. Aer will move to a standalone docs build and the main qiskit docs will have redirects from the old aer pages to the new build

1ucian0:
closed by https://github.com/Qiskit/qiskit-aer/pull/1589

PR #1596, opened 2022-08-31
Redirect aer docs to new import path

mtreinish:
### Summary

As part of the upcoming aer 0.11.0 release the project is moving to a
separate package/namespace and will no longer be part of `qiskit.*`.
This means the old documentation pages that existed for aer will no
longer be built because autodoc for the path will no longer be from
qiskit.providers.aer. To address this in the next metapackage release
this commit adds redirects from the existing aer documentation pages to
the new aer documentation.

### Details and comments

Related to #1508


mtreinish:
This is on hold until after aer 0.11.0 releases so the redirects have something to go to

mtreinish:
I've updated this PR to keep aer as part of the metapackage, I explained the rationale in: https://github.com/Qiskit/qiskit/pull/1596/commits/ac3169fbaf8f82b5f86b7868634ad657ba28230a for aer 0.12.0 we'll work on getting the standalone documentation in aer up to expectations. Then when aer isn't part of the metapackage anymore we'll be in a good place for updating the redirect to the standalone build

jakelishman (approved):
In principle this looks fine to me for once Aer 0.11 releases and is bumped into the metapackage.

jakelishman (approved):

PR #1597, opened 2022-09-07
fixes #1581: off by one docs version list (possible fix)

### LINKED ISSUE #1581: Stable version documentation doesn't show correct version of doc
### Informations

- **Qiskit version**:
- **Python version**:
- **Operating system**:

From 0.34, the stable release docs show the next version.
- https://qiskit.org/documentation/stable/0.34/index.html shows 0.35
- https://qiskit.org/documentation/stable/0.35/index.html shows 0.36
- https://qiskit.org/documentation/stable/0.36/index.html shows 0.37

image

Go to the links above

These links show go to the correct version of stable releases docs.

maximilien:
✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.

### Summary

I believe this will fix #1581 

### Details and comments

Docs list is off by one and I believe it's simply in the version list generation

I don't have a way to test it since CPLEX won't install on my Mac M1

I also don't see how to test docs either... unless you publish to the qiskit.org site which I would not want to do (obviously).

### DIFF
--- a/docs/versionutils.py
+++ b/docs/versionutils.py
@@ -94,7 +94,7 @@ def _get_version_list():
     if current_version_info[0] == '0':
         version_list = [
             '0.%s' % x for x in range(start_version[1],
-                                      int(current_version_info[1]) + 1)]
+                                      int(current_version_info[1]))]
     else:
         #TODO: When 1.0.0 add code to handle 0.x version list
         version_list = []


jakelishman (commented):
Thanks for the interest in contributing!

I don't think this fix is related - the list generation here just says which versions will appear in the "Previous releases" section, but the issue is that the documentation files that are in the  qiskit.org/documentation/stable/0.36 directory on our server are actually the 0.37 build.  The issue is to do with our deployment bash script (I'm _somewhat_ sure). I think I was going to sort that out a while back, then got distracted by other stuff going on in my life.

I think this change will just stop the latest release from appearing in the stable releases version (since `range` excludes the upper value, which is why there's a `+ 1` there).  We still do want that to appear (so there's a permalink available).

I've authorised the CI to run the docs build so we can check.

maximilien:
Thanks. How do I test the docs generation w/o pushing to Qiskit.org? I could link into the bash script... any other pointers appreciated.

jakelishman:
The docs build process is _super_ long for the metapackage, unfortunately.  If you've got `tox` installed (`pip install tox`), you can just run `tox -e docs`, but beware it may well take an hour.  It also will only build the current version of the docs - it won't build all the historical versions.  The current version inserts links to qiskit.org/documentation/stable/<version>, but the files that are there are just whatever was last pushed there by our deployment pipelines.  Ideally that's the _correct_ version, but clearly our deployment has a slight bug.

When we're checking docs PRs, our CI runs the build and publishes the build artifact.  [You can see the run here](https://github.com/Qiskit/qiskit/actions/runs/3004467135) (or you can do some clicking through "Checks" at the top of the PR page, then click the "Tests" heading.  If you scroll down, there should be an `html_docs` artifact you can download, which is the built documentation.

I made #1598, which I think is the actual underlying issue here (though I haven't properly had time to test it yet, or to re-deploy the broken stable links).

maximilien:
Thanks for the details. I will see what other docs or low hanging issues I can try to tackle next. Best.

PR #1598, opened 2022-09-08
Fix deploying new docs to old stable tag

### LINKED ISSUE #1581: Stable version documentation doesn't show correct version of doc

jakelishman:
### Summary

Previously, we relied on inferring the tag from `git describe`.  This has been somewhat unreliable in the past, seemingly not recognising the newest tag when triggered.  It is possible that the VM's git checkout is initialised before the tag is synced into the main repo, which might have caused this.

Instead, we extract the tag from the environment variable GitHub Actions sets to pass to the deployment script directly.  This is guaranteed to give the correct tag.

### Details and comments

Fix #1581.  Close #1597.  Actually, the online documentation won't be fixed until we run through the old stable versions and re-upload them, but this should fix the deployment script writing the first of a new minor version into the wrong place.

### DIFF
--- a/.github/workflows/docs_tag.yml
+++ b/.github/workflows/docs_tag.yml
@@ -25,5 +25,6 @@ jobs:
         encrypted_rclone_key: ${{ secrets.encrypted_rclone_key }}
         encrypted_rclone_iv: ${{ secrets.encrypted_rclone_iv }}
         QISKIT_DOCS_BUILD_TUTORIALS: 'always'
+      shell: bash
       run: |
         tools/deploy_documentation_tag.sh ${GITHUB_REF#refs/tags/}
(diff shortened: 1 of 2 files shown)


.github/workflows/docs_tag.yml
```diff
@@ -26,4 +26,4 @@ jobs:
         encrypted_rclone_iv: ${{ secrets.encrypted_rclone_iv }}
         QISKIT_DOCS_BUILD_TUTORIALS: 'always'
       run: |
-        tools/deploy_documentation_tag.sh
+        tools/deploy_documentation_tag.sh ${GITHUB_REF#refs/tags/}
```
mtreinish, line 30: Do we need to specify a shell here to use this syntax?

jakelishman: The shell will probably already be defaulting to bash (this script requires bash to at least be available), but it doesn't hurt to be explicit.  Done in 6b1f410d80.

mtreinish (approved):

PR #1606, opened 2022-09-21
Add s390x and ppc64le as tier 3 supported platforms

mtreinish:
### Summary

With the recent release of qiskit-aer 0.11.0 all the qiskit packages with compiled extensions now are publishing precompiled binaries for s390x and ppc64le. This commit updates the supported platform documentation to indicate that we support these platforms at tier 3, meaning we build and publish binaries at release time but do not have CI testing or infrastructure to test the binaries post build.

### Details and comments

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -577,11 +577,17 @@ platforms are best effort only.
 
 Tier 3 platforms are currently:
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
I actually thought we already had these here - I thought we'd already released Aer wheels with them once before!

PR #1608, opened 2022-09-27
Bugfix: docs header svg is now clickable

javabster:
### Summary
The svg for the header image on qiskit.org/documentation has hyperlinks for each title, but they are currently only clickable if a user opens the image separately. This PR updates the `index.rst` file to make the svg links clickable from the webpage (this file now uses the same html object used in `docs/contrubute_to_qiskit` to make the flowchart links clickable)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,5 @@
-.. figure:: images/qiskit_main.svg
-  :align: center
+.. raw:: html
+   :file: images/qiskit_main.svg
 
 ##############################
 Qiskit |version| documentation
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,13 @@
 .. raw:: html
-   :file: images/qiskit_main.svg
+
+  <div style="width: 100%; align: center">
+
+.. raw:: html
+  :file: images/qiskit_main.svg
+
+.. raw:: html
+
+  </div>
 
 ##############################
 Qiskit |version| documentation
--- a/docs/images/qiskit_main.svg
+++ b/docs/images/qiskit_main.svg
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   width="1137"
-   height="566.55072"
    viewBox="0 0 1137 566.55072"
    fill="none"
    version="1.1"


jakelishman (dismissed):
Yeah, I'm pretty sure this is the fix you need. This is what triggers the direct embedding of the SVG, rather than linking it through an `img` tag - I forgot the last time around.

javabster:
@jakelishman same i forgot too!

I'm just fixing up the formatting so it's aligned in the centre again then should be good to go

HuangJunye (changes requested):
Thanks for the fix! I can verify in the artifacts that the image is now clickable. 

However, the image overflow and it covers the TOC on the right. And simply scaling to 850px (max width of the main text) still causes problems if you rescale the window size. 

qiskit architecture svg pr

Can you try this instead (from [this answer](https://stackoverflow.com/questions/34777943/insert-clickable-svg-image-into-sphinx-documentation))? 

```rst
.. raw:: html

  <div class="figure align-center">
     <object data="myfile.svg" type="image/svg+xml"></object>
  </div>

```

HuangJunye:
Another comment is that the link for IBM provider is qiskit-ibm-runtime not qiskit-ibm-provider. Is that intended?

jakelishman:
Inserting as an "object" doesn't _feel_ right to me, but my intuition about HTML/CSS is pretty trash.  The SVG file having a hardcoded absolute width in its tag is what causing the resizing issues.  That should probably be blank (defaults to 100% of the container it's in) - I think if there's scaling issues with what constitutes the viewport, there's a `viewBox` attribute or something?

javabster:
thanks for the comments guys! Now I'm back from holiday I'll play around with the CSS and try get it sorted. Any suggestions always appreciated!

javabster:
@jakelishman @HuangJunye CSS fixed! ✨ ✨ ✨ 

Screenshot 2022-10-04 at 2 38 16 PM
Screenshot 2022-10-04 at 2 38 27 PM

jakelishman (approved):
Looks great!  Once #1610 merges to fix the CI issue you've got here, I think we're good, unless Junye spots anything else?

docs/index.rst
```diff
@@ -1,5 +1,13 @@
-.. figure:: images/qiskit_main.svg
-  :align: center
+.. raw:: html
```
HuangJunye, line 1: Just nitpicking. I think it would be good if we don't need to use three lines of `raw:: html`. According to [this article](https://vecta.io/blog/best-way-to-embed-svg), `<object>` is recommended if interactivity is required.
image

I haven't built the doc myself. Did what I suggest yesterday work?

```rst
.. raw:: html
  <div class="figure align-center">
     <object data="images/qiskit_main.svg" type="image/svg+xml"></object>
  </div>
```

HuangJunye (commented):
It works perfectly now! I think it would be good if we don't need to use three lines of ` .. raw:: html`. I also notice the image has change. Did you need to scale down the size of the image in order to fit with the main text container?

docs/index.rst
javabster: thanks Junye, yep I tried this but couldn't get the image to render 😢 I tried both object and img and whatever other suggestions google recommended, but the double raw html was the only solution that actually worked (in the limited time i had to look into this anyway haha)

javabster: and to answer your q about the image change it looks different because I had to remove the `width` and `height` props on the svg file itself, in order to get the html to override the sizing, which is why it looks different in github.  If you click the 'display source diff' button on the top right of the file diff view you can see the actual code i changed 😄 

Screenshot 2022-10-05 at 5 08 46 PM

HuangJunye (approved):

docs/index.rst
HuangJunye: LGTM. Thanks for the explanations!

PR #1613, opened 2022-10-13
use vanilla python/venv in getting started instead of anaconda/conda

### LINKED ISSUE #1611: Add a non-Anaconda option for getting started locally

The current ["Getting Started" docs](https://qiskit.org/documentation/getting_started.html#installation) recommends using Anaconda to set up a local environment. However, companies of over 200 people aren't allowed to use the open source (free) version, so it might be good to add another option to support those users (which I imagine would be a non-trivial number of folks). 

This would require an update to `qiskit/docs/getting_started.rst`. 

Perhaps something as simple as using `venv` to create the virtual environment rather than / in addition to `conda`? Could also do something more involved and make the setup not depend on a specific distribution (while still recommending Anaconda).

If this is something of interest, I'm happy to open a PR for it.

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

Fixes #1611 

### Summary

Updates the "Getting Started" locally section to use vanilla Python and `venv` instead of the Anaconda Python distribution and `conda`

### Details and comments

Updated the references to include links to downloading Python instead of Anaconda. 
Also added links to install Jupyter docs and the Python venv tutorial.
Also updated `conda` -> `pip` as appropriate in the "Install from Source" section (with the exception of for Windows, which [per the docs](https://github.com/Qiskit/qiskit-aer/blob/90a29293b41e3d8aa16b47c4ea93b55ec967ba7f/CONTRIBUTING.md#-dependencies--2) requires Anaconda).

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -18,33 +18,32 @@ going to run and install the packages.  There are three main ways to do this:
         evolving ecosystems, and sometimes when new releases occur in one or the other,
         there can be problems with compatibility.
 
-        We recommend installing `Anaconda <https://www.anaconda.com/download/>`__, a
-        cross-platform Python distribution for scientific computing. Jupyter,
-        included in Anaconda, is recommended for interacting with Qiskit.
+        You will need to `download Python <https://wiki.python.org/moin/BeginnersGuide/Download>`__
+        on your local system to get started. `Jupyter <https://jupyter.org/install>`__ is recommended for
+        interacting with Qiskit.
 
-        We recommend using Python virtual environments to cleanly separate Qiskit from
-        other applications and improve your experience.
-
-        The simplest way to use environments is by using the ``conda`` command,
-        included with Anaconda. A Conda environment allows you to specify a specific
-        version of Python and set of libraries. Open a terminal window in the directory
-        where you want to work.
-
-        It is preferred that you use the Anaconda prompt installed with Anaconda.
-        All you have to do is create a virtual environment inside Anaconda and activate the environment.
-        These commands can be run in the Anaconda prompt irrespective of Windows or Linux machine.
+        We recommend using `Python virtual environments <https://docs.python.org/3.10/tutorial/venv.html>`__
+        to cleanly separate Qiskit from other applications and improve your experience.
 
         Create a minimal environment with only Python installed in it.
 
         .. code:: sh
 
-            conda create -n ENV_NAME python=3
+            python3 -m venv /path/to/virtual/environment
 
         Activate your new environment.
 
         .. code:: sh
 
-            conda activate ENV_NAME
+            source /path/to/virtual/environment/bin/activate
+
+
+        Note: if you are using Windows, use the following commands in PowerShell.
+
+        .. code:: sh
+
+           python3 -m venv c:\path\to\virtual\environment
+           c:\path\to\virtual\environment\Scripts\Activate.ps1
 
 
         Next, install the Qiskit package.
@@ -53,7 +52,7 @@ going to run and install the packages.  There are three main ways to do this:
 
             pip install qiskit
 
-        If the packages were installed correctly, you can run ``conda list`` to see the active
+        If the packages were installed correctly, you can run ``pip list`` to see the active
         packages in your virtual environment.
 
         If you intend to use visualization functionality or Jupyter notebooks it is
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -18,33 +18,32 @@ going to run and install the packages.  There are three main ways to do this:
         evolving ecosystems, and sometimes when new releases occur in one or the other,
         there can be problems with compatibility.
 
-        We recommend installing `Anaconda <https://www.anaconda.com/download/>`__, a
-        cross-platform Python distribution for scientific computing. Jupyter,
-        included in Anaconda, is recommended for interacting with Qiskit.
+        You will need to `download Python <https://wiki.python.org/moin/BeginnersGuide/Download>`__
+        on your local system to get started. `Jupyter <https://jupyter.org/install>`__ is recommended for
+        interacting with Qiskit.
```
mtreinish, line 23: Sigh, I can clearly see why someone who isn't already well versed in Python or programming would want to use anaconda after comparing those two pages. The anaconda page is so much clearer on how to get going than that wiki page or from `python.org`. Reading the wiki page I'm also not really a fan of some of the recommendations they're making, like using a 32 bit executable on Windows. I've been looking for other pages on the python.org site and they all point to that wiki page or info dump download pages with a list of all the versions published. The closest alternative I've found is: https://www.python.org/about/gettingstarted/ but for the details that just links to the wiki page. 

That all being said I think this probably the best we'll get short of writing documentation ourselves or referring to some random tutorial not part of the Python docs. I've reached out to the Python website (pydotorg-www) mailing list with some proposed edits to the wiki page to hopefully improve the functionality.

mtreinish (dismissed):
I'm not super happy with the Python official install documentation, but I think what you've linked is the best we've got for now so lets go with it. We can always update it again if we have a better guide in the future (or want to write one ourselves).

mtreinish:
Oh, as I was going to merge I just noticed the details in the PR summary about installing from source section for windows requiring Conda. I don't know where that requirement in the documentation came from, but there isn't anything on windows I'm aware of that requires conda. I was able to install qiskit from source on my windows test vm with just mvsc and rust installed via pip without anything extra pretty easily. So we can probably just remove that line and update the commands to be pip based too.

docs/getting_started.rst
psschwei: If we wanted to make an opinionated choice on virtual environment tooling (i.e. using something other than `venv`), `pipenv` has very nice documentation: https://pipenv.pypa.io/en/latest/install/. That said, part of the point of this PR was to be less opinionated, so :shrug: 

I believe the same folks also wrote the Hitcherhiker's Guide to Python, which has some pretty nice install docs that we might want to use : https://docs.python-guide.org/starting/installation/

docs/getting_started.rst
```diff
@@ -344,24 +344,14 @@ going to run and install the packages.  There are three main ways to do this:
 
        .. dropdown:: Windows
 
-          On Windows you need to use `Anaconda3 <https://www.anaconda.com/distribution/#windows>`__
-          or `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`__ to install all the
-          dependencies.
+          3. Install a BLAS implementation; `OpenBLAS <https://www.openblas.net/>`__
+             is the default choice. Binaries can be downloaded from `GitHub <https://github.com/xianyi/OpenBLAS/releases>`__
```
mtreinish: Actually this can be skipped on aer too I think, aer vendors a binary version of openblas for windows so it's the only platform you don't need to install blas: https://github.com/Qiskit/qiskit-aer/tree/main/src/third-party

For windows it should just be the msvc c++ build/link toolchain, it uses conan to pull in binaries of other dependencies and that vendored openblas to work on a fresh windows environment.

psschwei: cool, I'll update to add the bit about the toolchain

psschwei: also bumping the msvc version, since there's a 2022 one out now

mtreinish: I don't think I've tested msvc 2022, I know in CI we're still using 2019. But if it's a problem that's an aer bug we should fix.

mtreinish (approved):
LGTM, thanks for the updates.

docs/getting_started.rst
psschwei: If I find a free cycle in the near future, I'll spin up a VM and test it out

PR #1616, opened 2022-10-16
Qiskit header image has the wrong link for Terra

cometta:
### Summary
fix link for svg for terra https://qiskit.org/documentation/apidoc/terra.html


HuangJunye:
@cometta Thanks for the contribution. Our original intention was to use the documentation homepage https://qiskit.org/documentation/ for Terra link because we are moving towards Qiskit == Qiskit Terra. But now I realize that the image is already in the home page and clicking on Terra just goes to the same page which is not very useful. I think it makes sense to change to https://qiskit.org/documentation/apidoc/terra.html.

What do you think @javabster @1ucian0?

1ucian0 (approved):
LGTM

1ucian0:
Thanks @cometta for your raising the issue and the PR. Welcome to the community!

PR #1617, opened 2022-10-18
convert code sh to text to avoid hightlighting in commands

1ucian0:
Showing snippets of code with highlight is usually a desired thing for readability. However, when terminal commands are highlighted with `sh`, then it is a bit distracting and does not add much:

Screenshot 2022-10-18 at 11 02 06

So, I suggest to use `text` in these cases.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -27,20 +27,20 @@ going to run and install the packages.  There are three main ways to do this:
 
         Create a minimal environment with only Python installed in it.
 
-        .. code:: sh
+        .. code:: text
 
             python3 -m venv /path/to/virtual/environment
 
         Activate your new environment.
 
-        .. code:: sh
+        .. code:: text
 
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -27,28 +27,28 @@ going to run and install the packages.  There are three main ways to do this:
...
             source /path/to/virtual/environment/bin/activate
 
 
         Note: if you are using Windows, use the following commands in PowerShell.
 
-        .. code:: sh
+        .. code:: text
 
            python3 -m venv c:\path\to\virtual\environment
            c:\path\to\virtual\environment\Scripts\Activate.ps1
```
jakelishman, line 46: This one should always have been `pwsh` presumably, but `text` seems fine to me too.

jakelishman (approved):
I personally don't mind the `sh` highlighting, but the Powershell ones are clearly wrong with the escape characters.  Happy to change it to text.

PR #1621, opened 2022-10-26
Reduce number of permutations for assembler benchmarks

mtreinish:
### Summary

When running the full benchmark suite the assembler benchmarks take a disproportionately large amount of time to run, especially given the relatively low current importance of the assemble() function. This was primarily due to the number of permutations of the benchmark we ran. At the time all those permutations were added to verify the scaling of the function because it was a potential bottleneck, but this is no longer a concern. This commit drops the parameter sweeps down so that we only run 2 combinations for assemble, basically a fixed sized circuit either by itself or as a batch of 100. This should drastically decrease the runtime cost of these benchmarks but still give us some data for the function moving forward.

### Details and comments

### DIFF
--- a/test/benchmarks/assembler.py
+++ b/test/benchmarks/assembler.py
@@ -22,9 +22,9 @@ from .utils import random_circuit
 
 
 class AssemblerBenchmarks:
-    params = ([1, 2, 5, 8],
-              [8, 128, 1024, 2048, 4096],
-              [1, 5, 10, 50, 100])
+    params = ([8],
+              [4096],
+              [1, 100])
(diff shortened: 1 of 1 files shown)


kevinhartman (approved):
Seems fine to me. Though would there be value in keeping the single qubit case?

mtreinish:
I didn't think so, the scaling of the assemble function should be linear with the number of instructions in the circuit.The qubit count there was mostly just a proxy for different sized circuits.

jakelishman (approved):
I think this represents a reduction in run-time of the benchmark suite by like 33% - 3h down to 2h or something like that. Should be a big help.

PR #1631, opened 2022-11-22
Update repo to use new qiskit sphinx theme

javabster:
### Summary
The latest version of the qiskit_sphinx_theme (1.10.0) includes 2 major new features:
- It replaces the existing top menu with the same top menu from qiskit.org
- It adds support for segment analytics (optional)

This PR removes the existing docs/templates/layout.html so that terra will stop overriding the qiskit_sphinx_theme and enable the new top menu to show. This PR also updates the `conf.py` to opt-in to analytics (details on what is tracked [here](https://github.com/Qiskit/qiskit_sphinx_theme#enable-qiskitorg-analytics))

I also updated the headings in the left sidebar, because some pages (e.g. tutorials) are specific to terra and not present in the new unified top bar. Similarly other pages (like contributing things) are relevant to all repos and are linked in the new top menu so don't need to be in the sidebar anymore.

Not sure if I did the env var setting the way you meant @mtreinish 😅


javabster:
my suggested new sidebar layout, based on the diataxis framework we are trying to follow going forward. 
Screenshot 2022-11-22 at 5 33 01 PM

docs/conf.py
```diff
@@ -251,6 +251,7 @@
 html_css_files = ['custom.css', 'gallery.css']
 html_favicon = 'images/favicon.ico'
 html_last_updated_fmt = '%Y/%m/%d'
+html_context = {'analytics_enabled': os.getenv('enable_analytics', False)} # enable segment analytics for qiskit.org/documentation
```
HuangJunye: Is it better to set this flag in the `docs.yml` instead of `conf.py` like you did in qiskit-nature: https://github.com/Qiskit/qiskit-nature/pull/979?

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
-  :caption: Frontmatter
-  :maxdepth: 2
   :hidden:
 
+  Overview <self>
   qc_intro
+  Getting Started <getting_started>
   intro_tutorial1
-  release_notes
+  Tutorials <tutorials/index>
```
HuangJunye: The tutorials sidebar item was not built for some reasons. It could be due to the wrong reference. I think it should be linked to the `tutorials.rst` in the root directory, not `tutorials/index.rst`
```suggestion
  tutorials
```

And I think whenever is possible, we can directly add the reference instead or adding links and the linked text can be directly pulled from the title of the linked page. In this case `tutorials.rst` has title `Tutorials` so `tutorials` in the `toctree` does the same as `Tutorials <tutorials>`. Then we don't need to worry about matching the linked text with the title text of the linked page.

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
-  Documentation Homepage <self>
 
 .. toctree::
-  :caption: Frontmatter
-  :maxdepth: 2
   :hidden:
 
+  Overview <self>
   qc_intro
+  Getting Started <getting_started>
```
HuangJunye: Similar to the comment below
```suggestion
  getting_started
```

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
   intro_tutorial1
-  release_notes
+  Tutorials <tutorials/index>
+  API Reference <apidoc/terra>
+  Circuit Library <apidoc/circuit_library>
+  Release Notes <release_notes>
```
HuangJunye:
```suggestion
  release_notes
```

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
   configuration
+  GitHub <https://github.com/Qiskit/qiskit-nature>
```
HuangJunye: This was probably a mistake
```suggestion
  GitHub <https://github.com/Qiskit/qiskit>
```

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
   :hidden:
 
+  Overview <self>
   qc_intro
+  Getting Started <getting_started>
   intro_tutorial1
-  release_notes
+  Tutorials <tutorials/index>
+  API Reference <apidoc/terra>
+  Circuit Library <apidoc/circuit_library>
```
HuangJunye: Should we keep circuit library in the side bar?

docs/index.rst
```diff
@@ -111,56 +111,22 @@ Interested in quantum hardware design?
...
-  :maxdepth: 2
   :hidden:
 
+  Overview <self>
   qc_intro
+  Getting Started <getting_started>
   intro_tutorial1
-  release_notes
+  Tutorials <tutorials/index>
+  API Reference <apidoc/terra>
```
HuangJunye, line 123: At the moment the meta package still contains qiskit-ear and qiskit-ibmq-provider. I think we should still add links to those pages. We can either add links to those API reference directly in the sidebar or add a new `apidoc/index.rst` and put `terra,rst`, `aer.rst`, `ibmq-provider.rst` in a `toctree`. I personally prefer the latter because it will not introduce another change to the sidebar when we remove these packages from the meta package.

HuangJunye (commented):
Thanks for working on this! I left a few comments for the sidebar items.

docs/conf.py
jakelishman: It's generally better to use an environment variable (as is done here) rather than hard-coding so it can be conditionally applied.  We don't necessarily want all CI builds / local builds to build in the analytics scripts.

That said, conventionally environment variables are ALL_CAPS, and we often start them with a "namespace", like `QISKIT_ENABLE_ANALYTICS` to avoid collisions in case setting it affects some other part of the stack.

docs/index.rst
jakelishman: > And I think whenever is possible, we can directly add the reference instead or adding links and the linked text can be directly pulled from the title of the linked page.

I'm not sure I agree with this, especially for the sidebar - the sidebar titles should always be short and don't necessarily need to match the main page title precisely, which might want to be a bit longer (since on their own pages, they're not in the same context).  I think decoupling them is actually a bit safer.

docs/index.rst
jakelishman: I don't see the problem with changing the sidebar, and I think hiding the actual API reference behind two layers of links is more annoying than not - it just makes it harder to click through from one method/class page back to the top-level API ref (for any package).

docs/conf.py
javabster: haha yeah i wasn't sure what convention to go with bc in the yaml there is both upper and lowercase ones 😅 I'll change to upercase

docs/index.rst
javabster: I think yes because it is an important feature of qiskit that we want to highlight more to users (as I understand)

docs/index.rst
javabster: I opted to remove them from the side bar because the same links are present in the top menu, and it also helps nudge people towards the idea that `qiskit = terra`

javabster:
thanks for the comments @jakelishman @HuangJunye ! I think i've addressed them all, lmk if theres anything else 😄

HuangJunye (approved):
Thanks for making the changes!

HuangJunye:
I am not tagging `automerge` in case @jakelishman and others want to have a final look.

jakelishman (approved):
I'm fine with this if you and Junye are - my comments were mostly just minor technical help rather than actual reviewing anyway!

PR #1635, opened 2022-12-02
Fix analytics environment in deployment

jakelishman:
### Summary

We were setting the `QISKIT_ENABLE_ANALYTICS` environment variable in the relevant workflow files, but building the documentation with `tox`. `tox` isolates its commands from the environment, except for variables that are explicitly passed through.

### Details and comments

cc: @javabster

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ commands =
 
 [testenv:docs]
 envdir = .tox/docs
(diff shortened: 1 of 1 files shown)


javabster (approved):
🚀 LGTM thanks for the speedy fix!

PR #1638, opened 2022-12-05
Fix toctree issues with documentation

mtreinish:
### Summary

In #1631 the docs were updated to use a new version of the qiskit_sphinx_theme and enable some new features it included. However, in doing this a few issues were introduced to the documentation. The biggest of which was the removal of the aer, circuit library, and the ibmq provider from the table of contents tree. Looking at the review comments on #1631 this was a consious choice to rely on hard coded navigation in the theme (i.e. the top bar). However, doing this breaks navigation in the sphinx documentation because it relies on the toctree to build the pathing for navigation from leaf pages in the toctree. For example, if you want to find the `QuantumVolume` class documentation currently you must search to find it because you can't get to it from any other page. Then when you find you're not able to get back to the parent page for the circuit library because it's an orphaned page outside the toctree. While aer and the ibmq provider get around this to a certain degree because you can get the per project index via the top bar (which I personally find non-obvious) navigation within that namespace is broken because they're not reachable from the root index. This commit adds back the missing references back to the root table of contents so navigation works again.

The other changes are relatively minor as it just fixes up the FAQ page which I mostly didn't realize existed prior to looking at the docs just now. Primarily it cleans up some irrelevant questions and also uses proper sphinx references to point to the install guide when needed.

### Details and comments


docs/faq.rst
```diff
@@ -61,9 +56,12 @@ Notebook.
 **A:** Qiskit depends on a number of other open source Python packages, which
 are automatically installed when doing ``pip install qiskit``. Depending on
 your system's platform and Python version, is it possible that a particular
-package does not provide a pre-built binary wheel for your system, and in those
-cases ``pip`` will attempt to compile the package from source, which in turn
-might require some extra dependencies that need to be installed manually.
+package does not provide a pre-built binaries  for your system. You can refer
+to :ref:`platform_support` for a list of which platforms are supported by
+Qiskit and those which may need an extra compiler. In cases where there are
```
kevinhartman:
```suggestion
package does not provide pre-built binaries for your system. You can refer
to :ref:`platform_support` for a list of platforms supported by Qiskit, some
of which may need an extra compiler. In cases where there are
```

1ucian0 (dismissed):
While I still think these links should be removed from the side menu. I opened #1643 and #1644 to follow up the discussion there.

In the meanwhile, let's reinstantiate them.

docs/index.rst
```diff
@@ -121,9 +121,12 @@ Interested in quantum hardware design?
   intro_tutorial1
   tutorials
   API Reference <apidoc/terra>
+  circuit_library
+  aer
+  ibmq-provider
```
mtreinish: I just realized I used the wrong links here: (this is why we really need to set -W on sphinx-build because this would have failed the build then)

```suggestion
  apidoc/circuit_library
  apidoc/aer
  apidoc/ibmq-provider
```

jakelishman (dismissed):
I'm fine with this - there's a technical need to get the landing pages back in a Sphinx-managed `toctree` - but one of the docs team (@1ucian0, @javabster, @HuangJunye, @agebbie) ought to look too.

docs/index.rst
```diff
@@ -120,12 +120,25 @@ Interested in quantum hardware design?
   getting_started
   intro_tutorial1
   tutorials
+
   API Reference <apidoc/terra>
+  apidoc/circuit_library
```
HuangJunye: Can we not add circuit library to side bar but add it to https://github.com/Qiskit/qiskit-terra/blob/main/docs/apidocs/terra.rst instead?

docs/index.rst
```diff
@@ -120,12 +120,25 @@ Interested in quantum hardware design?
...
+  apidoc/aer
+  apidoc/ibmq-provider
```
HuangJunye: The output looks confusing. 

image

I would prefer reverting to the old sidebar where API References has its own section.

image

Alternatively, which I prefer, is to add a new `apidoc/index.rst`, link it here and put `terra.rst`, `aer.rst`, `ibmq-provider.rst` in a `toctree`. as I [suggested](https://github.com/Qiskit/qiskit/pull/1631#discussion_r1031336448) in #1631

docs/index.rst
```diff
@@ -120,12 +120,25 @@ Interested in quantum hardware design?
...
 
+.. toctree::
+  :caption: Contribution Guidelines
+  :maxdepth: 1
+  :hidden:
+
+  contributing_to_qiskit
+  deprecation_policy
+  maintainers_guide
+
```
HuangJunye: Do we need to add back these links to sidebar as they can be accessed on the top menu and there are no other documentation depending on the these paths. 
image

HuangJunye (changes requested):
Thanks a lot for catching this issue. We apologize for not catching this issue in #1631. We would still prefer to remove these links from the sidebar if it still address the navigation issue. For that I suggested alternatives in the comments.

For contributing guidelines, the navigation is not an issue so I think we can remove them from the sidebar. Users can find these links on the top menu.

docs/index.rst
mtreinish: We can, but that requires a release because it will only get pulled into the documentation here on the next terra release. This is needed to fix it before the next terra release.

docs/index.rst
mtreinish: I'll add it as a toctree like what was there before.

docs/index.rst
mtreinish: We need them to be in the toctree so they're navigatable within sphinx. The problem is the top bar is completely divorced from the internal doc tree that sphinx builds (it's just hard coded links to the page's url). As a general rule you want every page to be in the tree so sphinx can build the complete information about it and also for local builds to be able to reason about the target url correctly.

As a smaller reason if we're ever going to be able to turn on `-W` in the builds (to treat warnings as errors) we'll need to ensure that every page included in the doc build is included in the toctree.

docs/index.rst
mtreinish: Done in: https://github.com/Qiskit/qiskit/pull/1638/commits/83748c368bc065965af7a3558c90251f10fcf01c

docs/index.rst
```diff
@@ -120,12 +120,31 @@ Interested in quantum hardware design?
...
 
+.. toctree::
+  :maxdepth: 1
+  :caption: API References
+  :hidden:
+
+  Qiskit <apidoc/terra>
+  Qiskit Circuit Library <apidoc/circuit_library>
+  Qiskit Aer <apidoc/aer>
+  Qiskit IBM Quantum Provider<apidoc/ibmq-provider>
```
HuangJunye:
```suggestion
  Qiskit IBM Quantum Provider <apidoc/ibmq-provider>
```

javabster:
I feel strongly that we should not implement the sidebar changes in this PR. The issues you are describing with the navigation seem pretty minor (ultimately, the user can still always get to where they need to go) whereas reverting the sidebar to the old style would be a major UI change for the user and would be a big step back in what we're trying to do with standardising docs and making things more terra focused. I could maybe see the argument for including the circuit library back in the sidebar as it is terra specific, but I don't think the Aer, provider and contributing links should go back.

Alternatively, I would suggest we make a small change to the sphinx theme (that we could get released before the next terra release) that would mean we don't display toctrees with the `:hidden:` directive - what do you think? The change needed to the theme is very small (https://github.com/Qiskit/qiskit_sphinx_theme/pull/167)

So in this case the `index.rst` would look something like this:

```rst
.. toctree::
   :maxdepth: 1
   :hidden:

   getting_started
   intro_tutorial1
   tutorials
   API Reference <apidoc/terra>
   release_notes
   configuration
   GitHub <https://github.com/Qiskit/qiskit-terra>
   faq

.. toctree::
   :maxdepth: 1
   :caption: Contributing
   :hidden:

   contributing_to_qiskit
   deprecation_policy
   maintainers_guide

.. toctree::
   :maxdepth: 1
   :caption: Other API References
   :hidden:

   apidoc/aer
   apidoc/ibmq-provider
```
This way we keep the new sidebar as is but we still have those other files as part of the toctree just not visible in the sidebar (if i understand the hidden directive properly)

If theres a reason this doesn't work then I still prefer @HuangJunye's  suggestion earlier (about moving the other api references under the api reference page)

mtreinish:
It's not a minor issue, navigation is fundamentally broken especially if you are building locally (the top bar doesn't work for builds outside of qiskit.org as the full url is hard coded into the links). There is no path for discovery of any content in the circuit library, aer, or ibmq outside the search bar. Then when you find the pages you're interested in there is no navigation between any of those pages as the hierarchy is broken without the toctree entries. As an aside, I also feel the new top bar is a big regression in the UX because none of the navigation is visible by default anymore. Before I could see the navigation content I cared about on every page of the documentation and now it's hidden behind a menu I have to know to click in the upper left. But, if the toctree was fixed and navigation were actually working I can just ignore this.

That all being said this has been broken for over a month now and it's quite frustrating to actually use the hosted documentation to get work done (so much so that I've stopped using the qiskit,org hosted documentation and am using a local build with this PR applied). So instead of bikeshedding on what the sidebar should look like where everyone clearly has a different opinion, we need to merge this to get this fixed ASAP and then you can change the sidebar to whatever makes you happy (as long as there aren't sections of the docs orphaned anymore)

javabster:
ok I understand you're frustrated but I think we can still find some middle ground here, and taking a bit of time now to understand the problem properly and put together a solution will prevent more confusion for users.

If i'm interpreting your frustrations correctly there are 2 aspects to the issue: a) that the sidebar/topbar UI has changed making it less intuitive (in your opinion) to find what you need and b) important pieces are missing from the toctree making the navigation broken. a) is highly subjective and we can debate it another time, and b) is the more pressing issue that needs to be resolved asap, and I believe we can resolve that without completely reverting to the old sidebar view.

So let's break it down even further:

1. **the circuit library.** As you pointed out this is missing from the toctree and not linked from either the sidebar or the top bar meaning the only way users can get there is through searching. I agree this is not great so we should add the circuit library back to the sidebar

2. **the contributing guidelines etc**. Users can still access them through the top menu and they are top level standalone pages so there isn't the navigation issue moving backwards as they anyway live directly under the docs homepage so there isn't the same toctree issue (unless I'm mistaken?). So i don't think these critically need to be moved back to the sidebar. As a side note these pages also aren't specific to terra (and the goal with sidebar is to only include terra specific links there and reduce duplication with the top menu)

3. **Aer and IBM provider**. As we are trying to move towards the narrative that `qiskit=terra` in an ideal world these would not be part of the sidebar (which should only contain terra specific pieces). Multiple design choices have been made throughout the docs revamp process specifically with the intention of making it clear to users that Aer and the IBM provider are not Qiskit, they are separate and should be accessed via the top menu. BUT I understand that currently because of how the metapackage is set up we have to include aer and ibm provider docs under the qiskit docs umbrella for the time being and removing them from the sidebar has caused issues with the toctree navigation. So for these I suggest we add them back to the sidebar (for now) but in a separate section where we make sure to signpost to users that these are separate. This also makes it easier for us to remove at a later date.

So ultimately I suggest the following:

```rst
.. toctree::
   :maxdepth: 1
   :hidden:

   getting_started
   intro_tutorial1
   tutorials
   API Reference <apidoc/terra>
   release_notes
   configuration
   GitHub <https://github.com/Qiskit/qiskit-terra>
   faq

.. toctree::
   :maxdepth: 1
   :caption: Contributing
   :hidden:

   contributing_to_qiskit
   deprecation_policy
   maintainers_guide

.. toctree::
   :maxdepth: 1
   :caption: Other API References
   :hidden:

   apidoc/aer
   apidoc/ibmq-provider
```

I think this is a reasonable compromise that solves the crux of the issue for the time being with minimal disruption to the users 😄

mtreinish:
That works fine for me except for the contributing guidelines, as I said before every document included in the sphinx build must be included in the toctree at some level. It doesn't need to be the top level root/index page but we must link to it from a toctree somewhere. The top bar by itself is not a solution in general because it is divorced from sphinx itself (meaning if you use other builders the documents are orphaned) and it also hard codes the url to `qiskit.org` which will not work for local `html` builds or anyone using the `html` builder output for something outside of our hosted build. This rule will also be enforced in the future during sphinx builds when we add `-W` to the `sphinx-build` call which we already do everywhere else except ibmq and here (because of ibmq).

HuangJunye:
> That works fine for me except for the contributing guidelines, as I said before every document included in the sphinx build must be included in the toctree at some level. It doesn't need to be the top level root/index page but we must link to it from a toctree somewhere. The top bar by itself is not a solution in general because it is divorced from sphinx itself (meaning if you use other builders the documents are orphaned) and it also hard codes the url to `qiskit.org` which will not work for local `html` builds or anyone using the `html` builder output for something outside of our hosted build. This rule will also be enforced in the future during sphinx builds when we add `-W` to the `sphinx-build` call which we already do everywhere else except ibmq and here (because of ibmq).

I suggest to add a new section "Interested in contributing to Qiskit?" and add the *non-hidden* `toctree` in the home page at the bottom. That way the pages are listed in a `toctree` and it won't show up in the sidebar. Do you think that's a good solution?

HuangJunye:
After discussions with @javabster and @mtreinish I updated the PR with the latest agreed sidebar

```rst

.. toctree::
   :hidden:

   Overview <self>
   qc_intro
   getting_started
   intro_tutorial1
   tutorials
   API Reference <apidoc/terra>
   Circuit Library <apidoc/circuit_library>
   release_notes
   configuration
   GitHub <https://github.com/Qiskit/qiskit-terra>
   faq

.. toctree::
   :caption: Contributing
   :hidden:

   contributing_to_qiskit
   deprecation_policy
   maintainers_guide

.. toctree::
   :caption: Other API References
   :hidden:

   apidoc/aer
   apidoc/ibmq-provider

```

The circuit library entry is added because the docs is built with Terra release not main branch (fixed in https://github.com/Qiskit/qiskit-terra/pull/9533). We can remove it after the next terra patch release [0.23.2](https://github.com/Qiskit/qiskit-terra/milestone/38) which is targeted to release on Feb 16.

javabster (approved):
LGTM 🚀

HuangJunye (approved):

PR #1642, opened 2022-12-08
Pin tox version used in CI to <4.0.0

mtreinish:
### Summary

The most recent tox release, 4.0.0, is a major rewrite of the internals of tox and several things behave quite differently. [1] This new release is causing CI jobs to fail as something in incompatible with our tox configuration (likely because it's using wheel builds instead of sdists). This commit pins the tox version we're using in CI to unblock things until we can update the tox configuration to be compatible with both the new and old versions of tox.

### Details and comments

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install Deps
-        run: python -m pip install -U tox setuptools virtualenv wheel pip
+        run: python -m pip install -U 'tox<4' setuptools virtualenv wheel pip
       - name: Install and Run Tests
         run: tox -e py
   lint:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         run: sudo apt-get install -y pandoc graphviz
       - name: Install python Deps
         run: |
-          python -m pip install -U tox
+          python -m pip install -U 'tox<4'
           python -m pip install -r requirements-dev.txt
```
1ucian0, line 54: why not adding it in `requirements-dev.txt`?

mtreinish: It ends up being a timing issue. Tox is normally used to install requirements-dev.txt as part of it's venv isolation: https://github.com/Qiskit/qiskit/blob/master/tox.ini#L41 so we would end up install tox in all of those venvs created by tox. I guess it's not a terrible waste to do that, but in the past we've just said you install tox manually and it manages everything for you.

1ucian0 (approved):

PR #1652, opened 2023-01-19
Use -d flag on sphinx-build to move .doctrees dir outside build

mtreinish:
### Summary

This commit updates our default sphinx-build invocation to include the -d flag to change the default output directory for the .doctrees dir (see: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d for details on the flag). The .doctrees dir is a local cache of pickle files for parsed source files and doesn't need to be included in the build. By moving the default location for this outside our build dir this will remove a lot of overhead from our uploaded docs as this cache can end up being quite large.

### Details and comments


jakelishman (approved):
Minor nitpicking/bikeshedding: perhaps it makes more sense to drop the doctrees somewhere in the build-output tree (e.g. `docs/_build/.doctrees`)?  I'm not bothered, though.

Eric-Arellano:
Still relevant?

mtreinish:
It is, we're uploading a bunch of unused and unnecessary data to COS without this. I didn't realize this never merged, I'm just going to tag as automerge now.

PR #1659, opened 2023-02-03
Enable manual trigger for Docs Publish workflow

HuangJunye:
### Summary

Currently the [Docs Publish workflow](https://github.com/Qiskit/qiskit/blob/master/.github/workflows/docs.yml) is trigged on push to `master` branch. However, there are scenarios where we would like to manual trigger Docs Publish even when there is no changes to the `master` branch such as:

1. There are changes to qiskit-tutorials repo and we want to rebuild the documentation page to reflect the latest changes.
2. There are new release of qiskit-sphinx-theme and we want to rebuild the documentation page to bring in changes introduced by the theme.

Currently we are re-running old Docs Publish jobs to as a work-around to trigger rebuild. However, that can be confusing to other collaborators if they see an old job got re-ran. Such as [the job below](https://github.com/Qiskit/qiskit/actions/runs/4058960426) I re-ran from Pump Meta commit to bring in changes from qiskit-sphinx-theme 0.11.2.

Screen Shot 2023-02-03 at 19 36 34


mtreinish (approved):
I'm fine with this, I don't feel super strongly that we need this, since we can retrigger the build via the ui as you mentioned in the summary. To be honest I highly doubt anyone looks at the docs deployment CI history unless they're debugging something specifically with the build system. But that being said adding this doesn't hurt so lets just add it

HuangJunye:
> I'm fine with this, I don't feel super strongly that we need this, since we can retrigger the build via the ui as you mentioned in the summary. To be honest I highly doubt anyone looks at the docs deployment CI history unless they're debugging something specifically with the build system. But that being said adding this doesn't hurt so lets just add it

I remember getting confused once when Luciano triggered the rebuild but I couldn't see it. I guess I am the only one that looks at the CI history LOL. Thanks for the approval!

PR #1662, opened 2023-02-07
Fix Python 3.7 support removal date (wrong year)

garrison:


mtreinish (approved):
Sigh, it wouldn't be a release without typos in the release notes :) Thanks for both catching this and fixing it. 

This is similarly incorrect in https://github.com/Qiskit/qiskit-terra/blob/stable/0.23/releasenotes/notes/0.23/prepare-0.23.0-release-0d954c91143cf9a4.yaml#L123 (which is only on the stable branch) If you don't mind updating it there in the source too.

PR #1669, opened 2023-02-15
Update introduction to Qiskit

ikkoham:
Import path was changed and AerSimulator is the latest simulator.  Unused numpy import was removed.

### Summary

I made the following changes:
- Import path was changed and AerSimulator is the latest simulator. 
- Unused numpy import was removed.
- Use black formatter.
- Change text code block to python code block.


docs/intro_tutorial1.rst
```diff
@@ -37,7 +36,7 @@ subsequent sections:
     circuit.cx(0, 1)
 
     # Map the quantum measurement to the classical bits
-    circuit.measure([0,1], [0,1])
+    circuit.measure([0, 1], [0, 1])
 
     # compile the circuit down to low-level QASM instructions
     # supported by the backend (not needed for simple circuits)
```
jakelishman: Let's remove the reference to "QASM" here - that's ancient as well.  Something like "Compile the circuit for the support instruction set and topology of the backend"?

docs/intro_tutorial1.rst
```diff
@@ -206,7 +203,7 @@ circuit will yield either the bit string 00 or 11.
 
 .. jupyter-execute::
 
-    simulator = QasmSimulator()
+    simulator = AerSimulator()
```
jakelishman, line 206: In the text just above this block it says we'll use "qasm_simulator".  I guess it probably means that we're using a statevector-based simulator - if so, can we be clearer about that?

jakelishman (commented):
Thanks for this - it's super easy for things like this tutorial to get lost in the cracks, so it's great to get it all updated.

docs/intro_tutorial1.rst
ikkoham: I guess what we are trying to say here is not that the method is `statevector` (in fact `AerSimulator` automatically selects the appropriate method), but that it loops over shots to compute the output.

javabster (approved):
this LGTM 🚀  The intro tutorials get some of the most traffic from new users so it's super important to keep these up to date - thanks for catching this!

1ucian0 (approved):

PR #1671, opened 2023-02-17
extension to add a prologue with a deprecation warning in ibmq pages

1ucian0:
This PR is the real https://github.com/Qiskit/qiskit-ibmq-provider/pull/1156. Prefixes all the pages related with `qiskit-ibmq-provider` with a warning message to a link to the migration guides.

Thanks @mtreinish for the help!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -360,6 +360,14 @@ def clean_tutorials(app, exc):
     tutorials_dir = os.path.join(app.srcdir, 'tutorials')
     shutil.rmtree(tutorials_dir)
 
+def deprecate_ibmq_provider(app, docname, source):
+    message = """.. warning::
+       The package ``qiskit-ibmq-provider`` is being deprecated and its repo is going to be
+       archived soon. Please transition to the new packages. More information in
+       https://ibm.biz/provider_migration_guide\n\n"""
+    if 'apidoc/ibmq' in docname or 'qiskit.providers.ibmq' in docname:
+        source[0] = message + source[0]
+
 # -- Extension configuration -------------------------------------------------
 
 def setup(app):
@@ -374,3 +382,4 @@ def setup(app):
     app.add_css_file('css/theme-override.css')
     app.connect('build-finished', clean_api_source)
     app.connect('build-finished', clean_tutorials)
+    app.connect('source-read', deprecate_ibmq_provider)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -365,7 +365,7 @@ def deprecate_ibmq_provider(app, docname, source):
        The package ``qiskit-ibmq-provider`` is being deprecated and its repo is going to be
        archived soon. Please transition to the new packages. More information in
        https://ibm.biz/provider_migration_guide\n\n"""
-    if 'apidoc/ibmq' in docname or 'qiskit.providers.ibmq' in docname:
+    if source[0] and 'apidoc/ibmq' in docname or 'qiskit.providers.ibmq' in docname:
         source[0] = message + source[0]
 
 # -- Extension configuration -------------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -365,7 +365,7 @@ def deprecate_ibmq_provider(app, docname, source):
        The package ``qiskit-ibmq-provider`` is being deprecated and its repo is going to be
        archived soon. Please transition to the new packages. More information in
        https://ibm.biz/provider_migration_guide\n\n"""
-    if source[0] and 'apidoc/ibmq' in docname or 'qiskit.providers.ibmq' in docname:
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -360,6 +360,14 @@ def clean_tutorials(app, exc):
     tutorials_dir = os.path.join(app.srcdir, 'tutorials')
     shutil.rmtree(tutorials_dir)
 
+def deprecate_ibmq_provider(app, docname, source):
+    message = """.. warning::
+       The package ``qiskit-ibmq-provider`` is being deprecated and its repo is going to be
+       archived soon. Please transition to the new packages. More information in
+       https://ibm.biz/provider_migration_guide\n\n"""
```
Eric-Arellano, line 367: Maybe use soft wrapping rather than hard wrapping? Right now, I believe this forces line breaks to occur where the code sets them. But that looks weird when you shrink or expand your screen -- we should rely on HTML and the browser to "soft wrap" the text for us.

This would look like this:

```suggestion
    message = (
       ".. warning::\n"
       "The package ``qiskit-ibmq-provider`` is being deprecated and its repo is going to be "
       "archived soon. Please transition to the new packages. More information in "
       "https://ibm.biz/provider_migration_guide\n\n"
```

I might be wrong here though - I'm not sure if Sphinx converts our hard wrapping to soft wrapping already.

docs/conf.py
```diff
@@ -360,6 +360,14 @@ def clean_tutorials(app, exc):
...
+    if source[0] and 'apidoc/ibmq' in docname or 'qiskit.providers.ibmq' in docname:
```
Eric-Arellano: What does `source[0]` mean? 

Is this infallible, that it's always a sequence with at least element 0? Or should it instead be `if len(source) >= 1`

Eric-Arellano (commented):
Neat!

docs/conf.py
1ucian0: from https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-source-read :
> The source argument is a list whose single element is the contents of the source file.

My understanding is that `len(source)` is always 1. I added `if source[0]` to avoid annotating pages without content (but it seems like not working tho).

docs/conf.py
jakelishman: Luciano's form shouldn't insert any hardwraps, and I think your suggestion has indentation errors, Eric.  The reStructuredText parser in Sphinx should join all source lines in a logical paragraph (with 1+ full blank between them) together.  I think the "use raw linebreaks" syntax is something like prepending all lines with a `|`, but I've _never_ used it.

Eric-Arellano: > Luciano's form shouldn't insert any hardwraps

Are you sure? If you look at the screenshot, the line wraps correspond exactly with the new lines in the code. Luciano, could you please try expanding and shrinking your screen with the current code to see what happens?

Unless Sphinx is doing special things to un-hard-wrap this, it would be hard wrapped because `"""` in Python treats new code lines as having a `\n` in between the text.

> The reStructuredText parser in Sphinx should join all source lines in a logical paragraph (with 1+ full blank between them) together. 

Ah, if true, then yeah this syntax would be good :)

>  I think your suggestion has indentation errors

Ah yeah, you're right.

HuangJunye: I think this is not a problem for Sphinx. The lines need to be separated by a new line to break. But that's in `.rst` file not in a `"""` string so let me double check the output artifact.

HuangJunye: It looks good.

image

You may already know, but you can download the artifact from CI here to see the build: https://github.com/Qiskit/qiskit/actions/runs/4204044350

image

HuangJunye: It's interesting that the line breaks exactly like the code in Luciano's screenshot ><

docs/conf.py
HuangJunye: Looks like the original code is ok. But I agree it looks weird.

docs/conf.py
Eric-Arellano: TIL. Thanks y'all!

docs/conf.py
jakelishman: I'm guessing here, but length-1 lists are sometimes used in Python interfaces to provide a very simple version of interior mutability for callback functions, instead of assigning the result of the call.  The string is immutable, but you can mutate the list to pass the result back to the caller.

It's an uncommon architecture because you can usually just return the value (or _values_, more commonly, since Python has ad-hoc compound return types (tuples)), but you can think of it as analogous to pass-by-mutable-reference semantics in other languages.

1ucian0: indeed. [The Sphinx core events](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events) change everything in-place.

Eric-Arellano (approved):

PR #1673, opened 2023-02-20
Move Sphinx extensions to `custom_extensions.py`

Eric-Arellano:
### Summary

This keeps `conf.py` more digestible. It makes it easier/clearer to add quality checks for our extensions, e.g. adding a `test_custom_extensions.py` file.

### Details and comments

No code was changed, only moved around.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,22 +18,17 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+import datetime
 import os
 import sys
+
 sys.path.insert(0, os.path.abspath('.'))
 
-import qiskit_sphinx_theme
+from docs import custom_extensions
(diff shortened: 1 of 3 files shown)


javabster (dismissed):
I think the changes in this PR makes sense, the `conf.py` is easier to read now and extending (😉 ) our list of extensions will be nicer in a separate file. One extra optional addition to this PR that would be nice would be to add a comment to each custom extension function just to briefly explain what it does. things like `deprecate_ibmq_provider` are self explanatory but others are less obvious

Eric-Arellano:
> One extra optional addition to this PR that would be nice would be to add a comment to each custom extension function just to briefly explain what it does. 

Good idea! Done.

HuangJunye (approved):
LGTM. The `conf.py` looks a lot cleaner now. Thanks for working on this.

PR #1674, opened 2023-02-20
Remove `cplex` and `tweedledum` dependencies

Eric-Arellano:
### Summary

Both of these aren't used anymore and can't be installed on Apple Silicon.

### Details and comments

Cplex was originally used by Qiskit Tutorials, but was removed from both the legacy_tutorials folder (deleted) and the 2 aqua tutorials we had in 2020.

Tweedledum is optional in Qiskit Terra. According to GitHub search, it's not used by any of the application repos.

With this PR, tox -e docs can successfully build the virtualenv 🎉


jakelishman (commented):
`tweedledum` is very likely required to build the full documentation, because it's required to import a certain subpackage of Terra (`qiskit.circuit.classicalfunction`), which is documented.  Tweedledum also may well not be used by name, but transitively through the `PhaseOracle` class, which I _think_ is either used / can be used by the Grover-search classes.

That module in Terra probably could be refactored to not need to import `tweedledum` at import time, which would let the docs build succeed on ARM Macs, but given that it's not maintained at all, we will need a complete replacement for it in the near future.  As far as I know, it's not a high priority despite that, because I don't think it's much-used functionality.

Eric-Arellano:
Matthew and I have been playing whack-a-mole, but so far it seems like the issues are coming from old release notes using `jupyter-execute`. This PR addresses that and explains the pros and downsides of getting rid of `jupyter-execute`: https://github.com/Qiskit/qiskit/pull/1676 - it's a "prework" PR for this one.

Eric-Arellano:
I removed the tweedledum part, since it needs more work: https://github.com/Qiskit/qiskit/pull/1676.

No need to block `cplex` on tweedledum. This at least gets some forward progress. Ready for review.

Eric-Arellano:
Bump when folks get a moment. This blocks a followup PR to get `tox -e docs` working on my M1

HuangJunye (approved):
https://github.com/Qiskit/qiskit/pull/1676 is merged now and the tests are passing so I am ready to approve. Pinging @jakelishman in case you have further comments.

Issue #1653, opened 2023-01-23
Release notes take a long time to load


king-p3nguin:
Current documentation takes a long time to load release notes. Splitting the release notes into separate pages for each version might improve loading times.

Eric-Arellano:
Relates to https://github.com/Qiskit/qiskit/issues/1504

jakelishman:
This issue is becoming obsolete as the current "release notes" file will become "legacy release notes" on Terra.  We may consider splitting it there, but for now, I'll close this issue as this repository is being sunset.

Issue #1504, opened 2022-05-30
Divide release note into seperate pages


HuangJunye:
### What should we add?

The [release note page](https://qiskit.org/documentation/release_notes.html) is getting very long. Would it make sense to divide it into separate pages for each release?

jakelishman:
This is to do with the metapackage, because the metapackage is what builds that release notes page.

Separate pages for _every_ release would make it very hard to look through at what's changed in a few old versions, but we could probably stratify it by some other criterion.  The trick is that this is all done very manually by copy-and-paste on releases, and we'd need to set up a bit more structure.  Possibly `0.2x`, `0.3x` and so on could work - for external linking purposes, we'd want to make sure that the release notes for a particular version always have a permanent URL (as best we can).

jakelishman:
This issue is becoming obsolete as the current "release notes" file will become "legacy release notes" on Terra.  We may consider splitting it there, but for now, I'll close this issue as this repository is being sunset.

PR #1676, opened 2023-02-21
Stop using jupyter-execute in release notes

Eric-Arellano:
### Summary

jupypter-execute has problems, as explained at https://github.com/Qiskit/qiskit-terra/issues/7661#issue-1137695552. One issue with `jupyter-execute` in release notes is it cannot handle deprecation changes well. For example, Terra made tweedledum optional; but due to old `jupyter-execute` blocks using those old features, https://github.com/Qiskit/qiskit/pull/1674 is failing.

For this reason, Terra has already removed Jupyter-execute from all its Reno release notes. But, those release notes have not been rebuilt to this meta repository. This PR ~applies those updates (technically, they're now mostly using `plot`, whereas this uses code blocks. That's fine because these are old release notes.)

### Details and comments


docs/release_notes.rst
```diff
@@ -3042,7 +3042,7 @@ New Features
   :meth:`.QuantumCircuit.draw` this label will be printed at the top of the
   ``barrier``.
 
-  .. jupyter-execute::
+  .. code-block:: python
```
Eric-Arellano: If we want to keep some `jupyter-execute` around, maybe a better approach is to only remove the blocks for Old™️ versions of Qiskit? For example, this block is from Terra 0.22 - so maybe it is actually Worth It to keep for now!

Whereas something from 2 years ago isn't worth the performance hit & deprecation issues

Eric-Arellano:
This seems fishy. Docs took 37 minutes vs 27 minutes on the latest `main` build at https://github.com/Qiskit/qiskit/actions/runs/4226514358/jobs/7340033415. @mtreinish any idea why this might be? I even reran the docs job in case caching is in play.

HuangJunye (commented):
The release note is getting very long time to load and build as noted in https://github.com/Qiskit/qiskit/issues/1653 and https://github.com/Qiskit/qiskit/issues/1504, though I am not entirely sure whether `jupyter-execute` is to be blamed for the long load time or just the fact that the release note is very long.

Nevertheless, this is one good step towards getting rid of `jupyter-execute`. I also agree that executable code for deprecated features are hard to handle. Perhaps we can put manual output inside the release note at the time of writing.

jakelishman (approved):
should be fine to change all of these - we did it on terra a little while back, when we completely removed the use of `jupyter-sphinx` from the repo.  We swapped to using matplotlib's `plot` directive for actual docs, but old release notes don't need to bother.

Eric-Arellano:
> we did it on terra a little while back, when we completely removed the use of jupyter-sphinx from the repo. We swapped to using matplotlib's plot directive for actual docs, but old release notes don't need to bother.

That's what I was thinking - Terra's Reno notes don't use jupyter-execute anymore, but these weren't updated. And that's causing pain like needing Tweedledum to build the docs.

Thanks for the review!

jakelishman:
yeah, they're "out of date" because there's no release-notes automation for the metapackage docs because reno isn't really designed for it.  We always just copy them in manually off a `reno report` as part of the metapackage release, so they get frozen in time

PR #1682, opened 2023-03-02
Minor import statement update 

SooluThomas:
### Summary

The [translation build](https://github.com/qiskit-community/qiskit-translations/actions/runs/4308564898/jobs/7514982619#step:5:164) in qiskit-translations repo errors with the error

```bash
Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/runner/work/qiskit-translations/qiskit-translations/docs/conf.py", line 27, in <module>
    from docs import custom_extensions
ModuleNotFoundError: No module named 'docs'
```

Though the reason for why it fails is a bit unclear, changing `from docs import custom_extensions` to `import custom_extensions` might fix the issue and can maintain a consistency.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,7 +24,7 @@ import sys
 
 sys.path.insert(0, os.path.abspath('.'))
 
-from docs import custom_extensions
+import custom_extensions
 from custom_directives import (IncludeDirective, GalleryItemDirective,
                                CustomGalleryItemDirective, CustomCalloutItemDirective,
                                CustomCardItemDirective)


jakelishman (approved):
For others: the context is that `from docs import ...` only works because it so happens that the docs builds of this repo happen to run Python from this repo's root, and the `sys.path.insert` line above enforces that `.` will be on the Python path.  Things like qiskit-translations, which also consume this build, do some more in-tree builds in part to avoid some potential extra files getting adding added to the path.

Sphinx always exec's `conf.py` from exactly in the source directory, which is why `import custom_directives` works in both cases.  For other consumers, it's better to do `import custom_extensions` as well, since that's now working-directory independent.

(Well, the ideal is maybe having `custom_extensions` and `custom_directives` as probably installable so they're available to other packages / builds as well, but within the current structure, Soolu's change should makes things safe for both this repo and qiskit-translations.)

Eric-Arellano (approved):
Thank you for fixing this!

PR #1688, opened 2023-03-10
Set up redirects for new Aer docs

### LINKED ISSUE #1686: Update Aer docs setup after #1685
https://github.com/Qiskit/qiskit/pull/1685#issuecomment-1462876107

Eric-Arellano:
### Summary

Closes https://github.com/Qiskit/qiskit/issues/1686.

### Details and comments

This also moves all redirect logic to a new `url_redirects.py` file. That makes `conf.py` much smaller and easier to read.


docs/custom_extensions.py
```diff
@@ -130,3 +133,32 @@ def deprecate_ibmq_provider(app, docname, source):
...
+        1) From `qiskit.org/documentation/stubs/qiskit.providers.aer` to
+           `qiskit.org/documentation/stubs/qiskit_aer`.
+        2) From `qiskit.org/documentation/stubs/qiskit_aer` to
+          `qiskit.org/documentation/aer/stubs/qiskit_aer`.
+
+    See https://documatt.gitlab.io/sphinx-reredirects/usage.html for how the returned dictionary
+    works.
+    """
+    result = {}
+    original_sources = (Path(__file__).parent / "aer_sources.txt").read_text()
```
Eric-Arellano: This `__file__` thing is so that the file can be correctly read both with Sphinx and our new unit test.

test/test_custom_extensions.py
```diff
@@ -0,0 +1,29 @@
...
+
+class CustomExtensionsTest(TestCase):
+    def test_determine_redirects_for_aer(self) -> None:
+        result = determine_redirects_for_aer()
+        num_lines_aer_sources = len(Path("docs/aer_sources.txt").read_text().splitlines())
+        self.assertEqual(len(result), num_lines_aer_sources * 2)
+
+        # Spot check some redirects.
+        self.assertEqual(
+            result["stubs/qiskit.providers.aer.AerError.html"],
```
Eric-Arellano: According to the docs, we're not supposed to have the `.html` at the end:

> Source (the key in redirects map) is a docname, i.e. document path without a suffix (an extension). Most Sphinx projects use .rst as extension. For example, a docname for the file index.rst is index, for agents/intro.rst is agents/intro, etc.
>
> Source may be non-existing document or existing document. If source does not exist, sphinx-reredirects creates new redirect .html file. For existing document, document’s .html file will be overwritten with redirect file. To select multiple existing documents, a [source wildcards](https://documatt.gitlab.io/sphinx-reredirects/usage.html#wildcard-syntax) can be used.

But, we were using `.html` suffix before. And it worked. So, I guess this is okay.

Eric-Arellano (commented):
@mtreinish how should we handle `qiskit_aer.extensions`? There are a bunch of entries in `aer_sources.txt`, but the module does not exist at https://qiskit.org/documentation/aer/apidocs/aer.html. Perhaps remove those entries from `aer_sources.txt`?

test/test_custom_extensions.py
mtreinish: I believe I had it without the `.html` before and it didn't work (because I remember reading the same thing in the docs). But we did this long enough ago I'm a bit fuzzy on the exact details.

Eric-Arellano: Weird. I see no reason to change it now

mtreinish:
> @mtreinish how should we handle `qiskit_aer.extensions`? There are a bunch of entries in `aer_sources.txt`, but the module does not exist at https://qiskit.org/documentation/aer/apidocs/aer.html. Perhaps remove those entries from `aer_sources.txt`?

I'd say lets just redirect them to the root of the aer documentation page. The `extensions` module was deprecated (for some time) and it was removed in the previous aer release 0.11.x, so anyone using that old path is probably getting a 404 now anyway. But I figure sending them to the new aer docs location is a bit kinder than a 404.

test/test_url_redirects.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class CustomExtensionsTest(TestCase):
+    """Test url_redirects.py."""
+    def test_redirects_for_aer(self) -> None:
+        """Test _redirects_for_aer()."""
+
+        result = _redirects_for_aer()
+        num_lines_aer_sources = len(
+            Path("docs/aer_sources.txt")
```
HuangJunye, line 16: I want to check whether my understanding is correct. The `aer_sources.txt` is a static snapshot of the api pages at Aug 2022 (https://github.com/Qiskit/qiskit/pull/1596). Do we need to update this file in this PR to reflect the latest api pages (if there are differences) so that all new pages will also get redirected?

HuangJunye (commented):
I like the approach to map the aer urls. Good job for noticing that there were two migrations with Aer, I would have forgotten about that. I left a comment about whether `aer_sources.txt` need to be updated. When that is addressed, I am ready to approve.

test/test_url_redirects.py
Eric-Arellano: No idea - @mtreinish or @1ucian0?

Eric-Arellano: We could land this now and then fix this in a follow up, if we decide it matters. In case we want to land this now before I'm OOO next week

mtreinish: For the most part I'm not sure how much has changed. It wouldn't hurt to add any new files that were part of aer 0.11.x which aren't present in this snapshot list. But also it's not as critical as getting this in place as a first step.

Eric-Arellano: @mtreinish do you remember how you generated the list?

In the meantime, we can land this PR. This Aer thing can be an update

mtreinish: IIRC, I just did an `ls` on the `docs/stubs` directory after building the aer docs and then just formatted the output as needed. The extra step we'll need here is that things like `apidoc/aer_provider.html`  will need to be redirected now too before they stayed the same so there wasn't anything to redirect.

mtreinish (approved):
LGTM, thanks for doing this

woodsp-ibm:
Does this still need to be published in some way? The live Qiskit Docs page as out there still points to this https://qiskit.org/documentation/apidoc/aer.html for Aer Docs which ends up in a 400 error.

HuangJunye:
> Does this still need to be published in some way? The live Qiskit Docs page as out there still points to this https://qiskit.org/documentation/apidoc/aer.html for Aer Docs which ends up in a 400 error.

~This link (https://qiskit.org/documentation/apidoc/aer) without `.html` works. Where did you find the link with `.html`?~ Actually it's not working. I manually triggered the rebuild and deploy [workflow](https://github.com/Qiskit/qiskit/actions/runs/4564704909). Let's see if that resolve the problem.

woodsp-ibm:
In the menu dropdown here
image

Ok, let see then...

woodsp-ibm:
I have just noticed that this is on the qiskit.org home page where I see this - the documentation drop down, if say I have first navigated, to Terra is correct

javabster:
I think the qiskit.org repo needs to bump it's version of the top menu component, I'll see if theres an issue open already

javabster:
issue opened for qiskit.org: https://github.com/Qiskit/qiskit.org/issues/3079\

HuangJunye:
Updating top menu will help, but the redirect should also work. I'll look into this tomorrow.

PR #1689, opened 2023-03-10
Replace pycodestyle with Black and Isort

Eric-Arellano:
### Summary

We use Black for all other projects.

We don't yet use isort from what I can tell, but there's been interest. 

### Details and comments

Also sets up a `pyproject.toml` with `build-backend`, which is best practice.

All files were manually generated other than `tox.ini` and `pyproject.toml`.


pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = ["setuptools", "wheel"]
```
jakelishman: I just learned today that including `"wheel"` here was just a historic suggestion from `setuptools`, and isn't actually required any more - [they even removed it from their docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

jakelishman (commented):
I personally don't care much about isort, but that cuts both ways - I also don't care if we use it.  I'm certainly happy with us running black on this repo.

I think we possibly want to exclude `test/benchmarks` from any formatting; doing so will change the asv versioning of those benchmarks, which will lose us all our accumulated history of them (well, I don't know _exactly_ what asv does when a benchmark version changes, but it definitely detects them).

mtreinish:
> I think we possibly want to exclude `test/benchmarks` from any formatting; doing so will change the asv versioning of those benchmarks, which will lose us all our accumulated history of them (well, I don't know _exactly_ what asv does when a benchmark version changes, but it definitely detects them).

The default behavior on asv is that the version number for a benchmark is a hash of the source code for the benchmark. When the version number changes it invalidates the old data and only shows results for the new version. I'm unclear if the data is still in the json results in the local asv db, but it does not show the data for the old versions anywhere in the webui if it does save the old data.

Eric-Arellano:
> I'm unclear if the data is still in the json results in the local asv db, but it does not show the data for the old versions anywhere in the webui if it does save the old data.

Meaning, it's okay to do this one-time formatting on those files?

mtreinish:
> > I'm unclear if the data is still in the json results in the local asv db, but it does not show the data for the old versions anywhere in the webui if it does save the old data.
> 
> Meaning, it's okay to do this one-time formatting on those files?

No, we shouldn't edit the benchmark code otherwise as @jakelishman said we won't see the accumulated history on https://qiskit.github.io/qiskit/ anymore. I was more just musing that the raw data might still be stored on the benchmark server, but regardless of that what matters in this case is that we continue to have the old data show on the dashboard because that's the primary interface for looking at the historical runtime data.

tox.ini
```diff
@@ -14,14 +14,23 @@ commands =
...
-  pycodestyle
+  black
+  isort
   pylint
   doc8
   ipython
 commands =
-  pycodestyle test
+  black --check docs/ test/ tools/ setup.py --extend-exclude=test/benchmarks/
+  isort --check docs/ test/ tools/ setup.py --extend-skip-glob='test/benchmarks/**'
```
jakelishman: I think we can put the `extend-exclude` stuff into the `pyproject.toml` file instead to make it easier to run these guys more manually as well (e.g. auto-formats from editors).  Certainly it's possible with `black` - I'm not so sure with `isort`.

test/test_url_redirects.py
```diff
@@ -8,46 +8,34 @@
...
-                "stubs/qiskit_aer.AerError.html"
-            ),
+            "https://qiskit.org/documentation/aer/stubs/qiskit_aer.AerError.html",
         )
         self.assertEqual(
-            result[
-                "stubs/qiskit.providers.aer.library."
-                "SaveAmplitudes.is_parameterized.html"
-            ],
+            result["stubs/qiskit.providers.aer.library." "SaveAmplitudes.is_parameterized.html"],
```
jakelishman: I think you've missed a split string here.

tox.ini
```diff
@@ -14,14 +14,23 @@ commands =
   pip check
   python -m unittest -v
 
+[testenv:fmt]
+deps =
+  black
+  isort
```
jakelishman: In both places, we should put pins on the `black` and `isort` versions to our chosen compatibility level.  Black's versioning scheme is year-based now, and they commit to no non-bug formatting changes within each calendar year, so we probably want `black~=23.0`.  I don't know what's appropriate with `isort`.

jakelishman (commented):
I have a couple of small comments, but in principle this seems fine to me.  We maybe want to consider putting `black` and `isort` with appropriate version pins in `requirements-dev.txt` for people who have their editors set to autoformat as they edit too.

Eric-Arellano:
Excellent feedback. Thanks Jake!

requirements-dev.txt
```diff
@@ -18,3 +18,7 @@ nbsphinx
 cvxpy
 networkx>=2.3
 sphinx-reredirects
+
+# Dev tooling
+black==23.3.0
```
jakelishman: Black claims to be stable enough that the correct pin is `black~=23.0`, and the next breaking change will be near the 1st of January 2024 (it's year-based major versioning), so we _maybe_ want to allow bugfix updates, but I'm not worried about it if you've got a reason to prefer the hard pin.

jakelishman (commented):
This looks fine to me.  There's apparently a merge conflict in the docs config now, but I'm assuming it's just the addition of a new line or two in a recent PR that needs formatting.

jakelishman (approved):
This seems fine to me in this situation.  This repo shouldn't be too much longer lived, but in the meantime it's nice to have sensible formatting in it.

PR #1690, opened 2023-03-10
Add `docs-clean` Tox command

Eric-Arellano:
### Summary

I keep getting confused between Qiskit vs Terra what `rm -rf` command to run. For example, in Terra, it's `docs/stubs`; in Qiskit, it's `stubs/`.

Now you can run `tox -e docs-clean` without having to know what to delete.

### Details and comments

This does not delete the Tox venv for docs. It is occasionally necessary to do if the venv is in a bad state, but the cost of recreating the venv is too high to always do delete it, given the low frequency. People can recreate the venv by using `tox -e docs -r`.

I will add an equivalent command to Terra in a follow up.


tox.ini
```diff
@@ -44,6 +44,13 @@ deps =
 commands =
   sphinx-build -b html {posargs} {toxinidir}/docs/ {toxinidir}/docs/_build/html
 
+[testenv:docs-clean]
+skip_install = true
+allowlist_externals =
+  rm
+commands =
+  rm -rf {toxinidir}/stubs/ {toxinidir}/docs/_build {toxinidir}/docs/apidoc {toxinidir}/docs/tutorials {toxinidir}/.tox/docs
```
jakelishman: I believe that the only things you need to clean are `docs/stubs` and `docs/_build`.  We definitely _don't_ need to clean `docs/apidoc` (that's the actual content that spawns all the documentation), and `.tox/docs` removes the tox venv, which seems like something we shouldn't do.  I think we should already largely be safe on that and we should probably fix anything that means you can't use the same environment to build the docs twice in a row (but I don't think there is anything). tox has a built-in command to rebuild an environment from clean anyway: `tox -r -e docs`.

Eric-Arellano: > We definitely don't need to clean docs/apidoc (that's the actual content that spawns all the documentation)

Our extensions already should be cleaning those up:

https://github.com/Qiskit/qiskit/blob/96f760475cb6ed9969c574dd6d79eec49165e930/docs/conf.py#L273-L274

But, that can fail. Like Sphinx being shut down ungracefully. I had those folders leftover on my machine when writing this PR.

This cleanup matters because without it, we will skip re-cloning the repo:

https://github.com/Qiskit/qiskit/blob/96f760475cb6ed9969c574dd6d79eec49165e930/docs/custom_extensions.py#L77-L81

That can mask issues, like that the versions have been bumped but what we downloaded is a prior version. Even if that 80% of the time isn't relevant, the purpose of this command is to get you in a totally clean state.

--

> and .tox/docs removes the tox venv, which seems like something we shouldn't do. I think we should already largely be safe on that and we should probably fix anything that means you can't use the same environment to build the docs twice in a row (but I don't think there is anything). tox has a built-in command to rebuild an environment from clean anyway: tox -r -e docs.

Yeah, that's intentional:

> This also deletes the Tox venv for docs. That is sometimes more aggressive than necessary, but there are some instances where it fixes bad state. If you want something less aggressive, manually run rm -rf.

The purpose of this command is to get you in a totally clean state. 

I could be convinced to remove the `.tox/docs` part, though. In part because Tox has good cache detection of when the venv needs to be recreated; although, it's worth noting it can't detect when new versions of packages are released on PyPI, such that a loose requirement like `sphinx-intl` would resolve to a new package than what is installed.

Eric-Arellano: Alright I got rid of the venv part. I realized it will be frustrating how much slower it makes `tox -e docs`, especially in Terra where the venv is much bigger.

It's less safe, but good enough.

jakelishman: oh crumbs sorry, I thought this was in the terra repo

jakelishman: yeah, you're definitely right we should clear `docs/apidoc` in the metapackage

tox.ini
```diff
@@ -44,6 +44,13 @@ deps =
 commands =
   sphinx-build -b html {posargs} -d {toxinidir}/docs/.doctrees {toxinidir}/docs/ {toxinidir}/docs/_build/html
 
+[testenv:docs-clean]
+skip_install = true
```
jakelishman, line 52: I think we want to be stronger than `skip_install` - we don't even need venv management, right?  I don't think we can convince tox to do that, but we can probably override it to have 0 dependencies to at least make the venv lighter.

I wonder if tox intends to allow this kind of Makefile-like usage; really we're deriving only overhead from having this in tox and not make, but it's better to have everything in one place.

Eric-Arellano: `skip_install` means "don't install the package itself". And then `deps` is empty. So, it does create a venv, but an empty venv.

I couldn't find a way in Tox's config to create no venv at all, unfortunately. But the first run isn't too bad with speed:

```
❯ rm -rf .tox/docs-clean
❯ time tox -e docs-clean
...
tox -e docs-clean  0.61s user 1.98s system 50% cpu 5.114 total
```

And then subsequent runs:

```
tox -e docs-clean  0.23s user 0.07s system 89% cpu 0.330 total
```

Eric-Arellano:
Bump

HuangJunye (approved):
Looks reasonable to me. This is a handy tox command to add. Thanks!

tox.ini
jakelishman: 2s isn't so bad, yeah.  It's a shame that there's no way to have no venv - it's just a waste of disk space, even if nothing else.

jakelishman (approved):

PR #1693, opened 2023-03-15
Use "Qiskit contributors" as authors in bibtex file

### LINKED ISSUE #934: Short version of Qiskit's bibfile
### Summary

Provide two `.bib` files to be used in references, a short one
```
Qiskit Contributors, “Qiskit: An open-source framework for quantum
computing,” 2019.
```
_additionally_ to the gargantuan one we currently have.

### Reasoning

Including the bibfile of Qiskit in a paper easily takes up half a column and includes GitHub usernames. Practically, many papers just put the name of the first author to get
```
H. Abraham et al., “Qiskit: An open-source framework for quantum
computing,” 2019.
``` 
This could be improved, because 
1. It's an extra manual step to remove all the authors.
2. Putting `Qiskit Contributors` instead of a single name, at least in my opinion, reflects the community effort of Qiskit better.

Cryoris:
### Summary

Replace the author list in the `Qiskit.bib` file by "Qiskit contributors". Previously discussed in #934.

### Details

Note that if this is merged we have to update [`Qiskit/qiskit-bot`](https://github.com/Qiskit/qiskit-bot/blob/master/qiskit_bot/release_process.py#L33-L55) to not overwrite the bibfile upon release.

#### Current status

To cite Qiskit we currently provide the [`Qiskit.bib`](https://github.com/Qiskit/qiskit/blob/master/Qiskit.bib) file, which has an autogenerated author list including all account names appearing in any commit to one of Qiskit Terra, Aer, Ignis, IBMQ Provider, the applications or the meta package.

This approach has several shortcomings:

1. Including the bibfile of Qiskit in a paper easily takes up more than a page and includes GitHub usernames. For example, the current first author is `A-tA-v`, so even if the journal would automatically truncate, we would be cited as `A-tA-v et al. Qiskit: ....`. There are a lot of these cases in the full list.
2. Qiskit is a community effort; using `Qiskit contributors` as author would reflect that explicitly.
3. Qiskit is a dynamic ecosystem, some central repositories, like Qiskit IBM Runtime or Qiskit Experiments, are not included.
4. In practice, many (most?) papers manually truncate the author list anyways to avoid taking up 1 page for one citation.

#### Proposal

By using `Qiskit contributors` as author name, which is a commonly used approach for large software projects or collaborations (e.g. CERN),  we could resolve all of the above issues. On zenodo, we can still reference the [author list](https://github.com/Qiskit/qiskit/blob/master/AUTHORS) or just reference GitHub's contributor list.

One concern raised in #934 is that by using `Qiskit contributors`, contributors would not get credit if Qiskit is cited. However, as long as the proper zenodo link is included (as per our bib file), the citation is correctly detected. To have Qiskit appear in e.g. Google Scholar, the article can always be added manually -- instead of the current opt-out we would move towards an opt-in. Given that many contributors don't have Qiskit in their Google scholar account anyways, this seems like a small drawback while gaining a lot of advantages.

#### In a nutshell

**Pro**

* clean and concise citation (which many papers currently do manually)
* no problem with usernames as authors 
* reflects the community effort and independent of changes/renamings in the Qiskit ecosystem

**Con**

* opt-in for the citation on citation trackers (like Google Scholar) instead of opt-out


mtreinish (commented):
I think we can also delete the authors list and the corresponding script to generate that file. There isn't really anything useful it provides anymore without the bibtex file

mrossinek:
> I think we can also delete the authors list and the corresponding script to generate that file. There isn't really anything useful it provides anymore without the bibtex file

I agree with this :+1: Instead of linking to this file as suggsted in the PR description, we could do it similar to how [rustworkx does it](https://zenodo.org/record/7352046) and refer to the `Contributors` graph page on Github. Yes, this would currently limit the scope to the core Qiskit repo (i.e. https://github.com/Qiskit/qiskit-terra/graphs/contributors) but more and more Qiskit ecosystem components are setting up their own DOIs via Zenodo (e.g. rustworkx, qiskit-metal) so I do not see this as a big drawback (plus it helps with the "what is Qiskit" question if we encourage submodules to be cited explicitly, too, as Qiskit would only refer to core also in terms of citation purposes)

nonhermitian:
To see why this is most helpful see Ref. 49 of this paper: https://arxiv.org/abs/2303.07660

mtreinish (commented):
The other place you need to update is the https://github.com/Qiskit/qiskit#authors-and-citation which refers to the `AUTHORS` file still

Cryoris:
> The other place you need to update is the https://github.com/Qiskit/qiskit#authors-and-citation which refers to the AUTHORS file still

Removed in 67449bec1bb802de295948cb4eba8894d52d10eb

mtreinish (approved):
This LGTM, cleaning this up will make citing qiskit a lot easier for people.

PR #1694, opened 2023-03-16
Fix `tox -e docs` for macOS

Eric-Arellano:
### Summary

`tox -e docs` now works on my M1.

### Details and comments

We now say to install `.[all]` in two places: before installing the package itself when just installing `deps`, and then again when installing the package itself. Tox uses the same venv for both times, so that essentially results in a no-op in the second stage. This is similar to https://github.com/Qiskit/qiskit-terra/pull/9758.


Eric-Arellano:
Bump

HuangJunye (approved):
LGTM.

PR #1715, opened 2023-04-10
Remove duplicated HTML templates

Eric-Arellano:
### Summary

The templates `stable_versions.html` and `versions.html` are identical to qiskit_sphinx_theme's version in 0.10. There's no benefit to duplicating them here.

Also, those templates are no longer used in the theme version 0.11.

### Details and comments

Also fixes our `.gitignore` and `docs-clean` command to handle the `plot_directive` output that happens when building docs locally now.

Removes the line `css/theme-override.css`, which we now instead set via this repo's `conf.py` pointing to the file `_static/custom.css`.


requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 qiskit-ibmq-provider[visualization]
 numpy>=1.17
 Sphinx~=4.0
-qiskit-sphinx-theme>=1.10.0
+qiskit-sphinx-theme~=1.10.3
```
Eric-Arellano, line 4: Now, we'll explicitly upgrade to new versions of the theme. The theme sometimes has substantial (possibly breaking) changes between minor releases. 

So, it's more stable for us to consciously upgrade to the new theme in a dedicated PR that shows how it renders. Rather than upgrading whenever we happen to release the newest version of the theme.

jakelishman (approved):
In general it would be best if the theme doesn't make frequent breaking changes (in terms of necessary configuration of downstream repos), since every time that happens a whole bunch of repositories have to update which just uses up everyone's time, but I'm totally on board with pinning the version of packages used for the rendered documentation.

Eric-Arellano:
> In general it would be best if the theme doesn't make frequent breaking changes (in terms of necessary configuration of downstream repos)

+1, we agree with this.

PR #1728, opened 2023-05-05
Fix opflow migration link in release notes

woodsp-ibm:
### Summary

Corrects the short link for the opflow migration in release notes

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -58,7 +58,7 @@ the :mod:`qiskit.algorithms` module to be based on the computational
 existing users to migrate to the new workflow:
 
   * ``QuantumInstance`` migration guide: https://qisk.it/qi_migration
-  * ``Opflow`` migration guide: https://qisk.it/qi_migration
+  * ``Opflow`` migration guide: https://qisk.it/opflow_migration
   * Algorithms migration guide: https://qisk.it/algo_migration
 
 OpenQASM2 improvements


Eric-Arellano (approved):

mtreinish (approved):
Thanks for fixing this!

docs/release_notes.rst
```diff
@@ -58,7 +58,7 @@ the :mod:`qiskit.algorithms` module to be based on the computational
 existing users to migrate to the new workflow:
 
   * ``QuantumInstance`` migration guide: https://qisk.it/qi_migration
-  * ``Opflow`` migration guide: https://qisk.it/qi_migration
+  * ``Opflow`` migration guide: https://qisk.it/opflow_migration
   * Algorithms migration guide: https://qisk.it/algo_migration
```
HuangJunye, line 62: Now that we have all three migrations, can you use sphinx cross-reference instead of `qisk.it` links?

HuangJunye (commented):
Thanks for catching this! Now that we have all three migration guides merged into terra and [the metadoc pulls the migration guides](https://github.com/Qiskit/qiskit-metapackage/pull/1664), can you use sphinx cross-reference instead of `qisk.it` links?

mtreinish:
I think for now lets keep using the short links. I think the reason we went this direction originally is that we can update them to the stable version of the docs if we want to drop the migration guides from the docs tree in a future qiskit release. But either way this is just fixing an obvious copy/paste error in the release notes and we can discuss the exact form of the links in a follow up PR or issue.

PR #1731, opened 2023-05-12
Remove unused custom CSS files and Sphinx directives

Eric-Arellano:
No one was using the custom directives `galleryitem`, `customgalleryitem`, `customcarditem`, and `includenodoc`. (I searched all Qiskit repos using ripgrep locally.)

After removing this, it's safe to remove `sphinx-gallery` because it's not used. That allows removing `gallery.css`, which is meant to override `sphinx-gallery`'s default styling.

This also removes `custom.css`. There is no `.toggle` element anymore.


docs/custom_directives.py
```diff
@@ -10,19 +10,12 @@
...
 
-import sphinx_gallery
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
 from docutils.statemachine import StringList
 
-try:
-    FileNotFoundError
-except NameError:
-    FileNotFoundError = IOError
```
Eric-Arellano, line 24: Legacy Python 2 code, I believe

javabster:
what are we using for tutorials if we're not using sphinx-gallery? I always thought that was the package that was used to lay out all the boxes on the tutorials page, unless I'm mistaken?

Eric-Arellano:
> what are we using for tutorials if we're not using sphinx-gallery? I always thought that was the package that was used to lay out all the boxes on the tutorials page, unless I'm mistaken?

Almost every project in Qiskit is using `nbsphinx`, which also has a tutorial gallery feature. For example, we use that in our example_docs/. 

Meanwhile, not one Qiskit project is using sphinx-gallery anymore. I suspect that you all migrated from sphinx-gallery to nbsphinx in the past.

javabster (dismissed):
LGTM 🚀

javabster (dismissed):

javabster (approved):

PR #1735, opened 2023-05-16
Remove jupyter-execute from docs

Eric-Arellano:
Similar to https://github.com/Qiskit/qiskit-terra/pull/9346. For us to migrate metapackage docs to Terra, we will need to stop using jupyter-execute.

This introduces a small regression to intro_tutorial1, which is already known with https://github.com/Qiskit/qiskit-terra/issues/10614 and a problem for Terra docs. It adds a stray `()` to the docs.


docs/qc_intro.rst
```diff
@@ -163,13 +163,7 @@ which supports a rich feature set of operations and can send your circuits to a
 With only a few lines of code, it is possible to construct complex circuits like the
 one above.
 
-.. jupyter-execute::
-   :hide-code:
```
Eric-Arellano, line 167: Because we were hiding code, the only benefit of using `jupyter-execute` was testing the code.

docs/intro_tutorial1.rst
```diff
@@ -53,9 +54,10 @@ subsequent sections:
     print("\nTotal count for 00 and 11 are:", counts)
 
     # Draw the circuit
-    circuit.draw()
+    circuit.draw("mpl")
 
-.. jupyter-execute::
+.. plot::
+    :include-source:
```
mtreinish, line 61: I don't think this will be a drop in replacement (I actually expect this to fail), the `plot` directive just calls. There isn't a shared context between plot calls by default. Jupyter execute spawned a separate kernel for each document in the sphinx build, but for plot by default it doesn't share context. However, I just looked at the docs and there is a `:context:` option we can use to do this: https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html#options

```suggestion
.. plot::
    :include-source:
    :context:
```

We'll have to do this with the call before this too.

docs/intro_tutorial1.rst
```diff
@@ -201,7 +204,8 @@ Terra.
 To simulate this circuit, you will use the ``AerSimulator``. Each run of this
 circuit will yield either the bit string 00 or 11.
 
-.. jupyter-execute::
+.. plot::
+    :include-source:
```
mtreinish, line 211: Similarly this will fail because there isn't shared context

docs/intro_tutorial1.rst
```diff
@@ -229,7 +233,8 @@ Qiskit provides `many visualizations <apidoc/visualization.html>`__,
 
 including the function ``plot_histogram``, to view your results.
 
-.. jupyter-execute::
+.. plot::
+  :include-source:
```
mtreinish, line 241: Same here

mtreinish (commented):
Overall I'm in favor of doing this, but I left some comments inline. `plot` isn't a drop in replacement for `jupyter-execute` as it doesn't have a shared context like `jupyter-execute` did so the code blocks either need to be modified or we can try using the  `context` option.

Eric-Arellano:
Much better! Thanks Matthew. FYI I had to add `close-figs` because otherwise we render the same diagrams multiple times.

This should now be ready.

Screenshot 2023-05-17 at 1 53 00 PM

mtreinish (approved):
LGTM, thanks for the quick update

PR #1742, opened 2023-05-22
Remove sphinx-redirects in favor of Cloudflare config

### LINKED ISSUE #1738: Move Aer redirects to Cloudflare
Right now, we are using the extension `sphinx_redirects` to redirect the old Aer docs to the new style. (Note that this before the Ecosystem migration):

https://github.com/Qiskit/qiskit-metapackage/blob/d81654d4caf0515989e79079c2f8ed94395220e0/docs/url_redirects.py#L124-L151

But sphinx_redirects does not properly set its Sphinx config if it can be parallelized. So it causes `-j auto` to fail, which makes the docs build much slower. That will be an issue when we migrate the metapackage docs to Terra.

Instead, Matthew T. had the simpler idea to move these redirects into the Cloudflare config for qiskit.org. That is already how we did the ecosystem/ redirects.

Eric-Arellano:
Closes https://github.com/Qiskit/qiskit-metapackage/issues/1738. We have moved these redirects to CloudFlare. For example, the bogus URL http://qiskit.org/documentation/tutorials/optimization/my_tutorial.html correctly goes to http://qiskit.org/ecosystem/optimization/tutorials/index.html


Eric-Arellano (commented):
Blocked by merging and deploying the internal change for our qiskit.org config. But this PR is ready for review, otherwise.

mtreinish (approved):
LGTM, I confirmed the equivalent redirects are in place int he cloudflare configuration so we no longer need to do this at the sphinx build level.

PR #1743, opened 2023-05-22
Fix Sphinx docs issues with extlinks

Eric-Arellano:
We've been getting this error in our docs, which blocks us from enabling `-W` (error on warnings):

```
WARNING: extlinks: Sphinx-6.0 will require a caption string to contain exactly one '%s' and all other '%' need to be escaped as '%%'.
```


docs/conf.py
```diff
@@ -133,9 +134,11 @@
 # -- Configuration for extlinks extension ------------------------------------
 # Refer to https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html
 extlinks = {
-    "pull_terra": ("https://github.com/Qiskit/qiskit-terra/pull/%s", "#"),
```
Eric-Arellano, line 131: Unused

docs/conf.py
```diff
@@ -133,9 +134,11 @@
...
-    "pull_aer": ("https://github.com/Qiskit/qiskit-aer/pull/%s", "#"),
-    "pull_ibmq-provider": ("https://github.com/Qiskit/qiskit-ibmq-provider/pull/%s", "#"),
+    "pull_aer": ("https://github.com/Qiskit/qiskit-aer/pull/%s", "qiskit-aer #%s"),
+    "pull_ibmq-provider": (
+        "https://github.com/Qiskit/qiskit-ibmq-provider/pull/%s",
+        "qiskit-ibmq-provider #%s",
+    ),
```
Eric-Arellano, line 136: These are used in release_notes.py, like this:

```
 Refactor thread management :pull_aer:`50`
```

This changes the text that we will render. Before, `#50`. Now, `qiskit-aer #50`.

docs/conf.py
```diff
@@ -217,6 +220,6 @@
 def setup(app):
     custom_extensions.load_api_sources(app)
     custom_extensions.load_tutorials(app)
-    app.setup_extension("versionutils")
+    versionutils.setup(app)
```
Eric-Arellano: This change is to avoid Sphinx complaining about missing parallelism options.

docs/maintainers_guide.rst
```diff
@@ -250,19 +250,19 @@ The module-level docstring
 
    For example, as in the previous dagcircuit module example, the
    contents of the module docstring for ``qiskit/dagcircuit/__init__.py`` would
-   be::
+   be like the below (remove the double backticks if copying from the RST file)::
```
Eric-Arellano: I spent more time than I'd like to admit trying to get the below to work. Outside of outright deleting this all, this is the best I could come up.

Eric-Arellano: Hm this is how it renders. Not great. Thoughts on what to do?

Screenshot 2023-05-24 at 12 35 00 PM

Eric-Arellano:
I removed the controversial parts of this PR because I'm going to open a dedicated PR for it. It's bigger surgery. This PR is ready to go and has no controversy.

docs/conf.py
HuangJunye: I didn't know this extension exist. It's quite handy. We should use it more.

docs/conf.py
HuangJunye: I found many release notes refer to terra issues but the links are normal links. See https://qiskit.org/documentation/release_notes.html#bug-fixes

Screenshot 2023-05-31 at 15 59 02

For example ``#9798 <https://github.com/Qiskit/qiskit-terra/issues/9798>`` in:

https://github.com/Qiskit/qiskit-terra/blob/main/releasenotes/notes/0.24/fix-9798-30c0eb0e5181b691.yaml

HuangJunye: If we keep others, I think we should keep `pull_terra` and advertise the usage of this plugin. If not, perhaps we can change where `pull_aer` and `pull_ibmq-provider` are used to normal links and get rid of the extlinks altogether.

Eric-Arellano: Sure, added back.

1ucian0 (approved):

PR #1744, opened 2023-05-23
Fix URLs used in stable docs deploys

Eric-Arellano:
I missed this in https://github.com/Qiskit/qiskit-metapackage/pull/1740. We renamed the option from `content_prefix` to `docs_url_prefix`.

### DIFF
--- a/docs/versionutils.py
+++ b/docs/versionutils.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 
 
 def setup(app):
+    app.connect("config-inited", _extend_html_context)
     app.add_directive("version-history", _VersionHistory)
 
 
--- a/tools/deploy_documentation.sh
+++ b/tools/deploy_documentation.sh
@@ -21,7 +21,7 @@ sudo apt-get install -y ./rclone.deb
 RCLONE_CONFIG_PATH=$(rclone config file | tail -1)
 
 # Build the documentation.
-tox -edocs -- -D content_prefix=documentation -j auto
+tox -edocs -- -j auto
 
 echo "show current dir: "
 pwd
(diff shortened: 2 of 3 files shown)


tools/deploy_documentation.sh
```diff
@@ -21,7 +21,7 @@ sudo apt-get install -y ./rclone.deb
 RCLONE_CONFIG_PATH=$(rclone config file | tail -1)
 
 # Build the documentation.
-tox -edocs -- -D content_prefix=documentation -j auto
+tox -edocs -- -j auto
```
Eric-Arellano, line 24: Now set here, so no need to configure via CLI override:

https://github.com/Qiskit/qiskit-metapackage/blob/86a29b4b4b744417c9f478d2252aa735db8bd568/docs/conf.py#L43

jakelishman (dismissed):
Seems odd to me to have this inconsistency in where the documentation prefix is defined ("as an input to sphinx-build" seemed more logical than "universal conf file" to me, since it's inherently tied to one particular deployment), but that's not a particularly big deal.  Good to get the options fixed.

Eric-Arellano:
> Seems odd to me to have this inconsistency in where the documentation prefix is defined ("as an input to sphinx-build" seemed more logical than "universal conf file" to me, since it's inherently tied to one particular deployment)

I'd be fine with either way. 

The important detail is that qiskit_sphinx_theme will error if versions or translations are used and `docs_url_prefix` isn't set. So, I suppose then it would be safer to not set it in `conf.py` because it forces every call site of `sphinx-build` to reason about it, rather than not realizing they're using the default.

docs/versionutils.py
```diff
@@ -25,6 +25,7 @@
 
 
 def setup(app):
+    app.connect("config-inited", _extend_html_context)
```
Eric-Arellano, line 28: My bad. This results in the Previous Releases feature not showing up.

HuangJunye (approved):

PR #1750, opened 2023-06-02
Qiskit Partners program is sunset

1ucian0:
Update the README file to remove information about the Qiskit Partners program, as it is sunset.


Eric-Arellano (dismissed):
Ha, I'm working on very similar change right now and I wasn't sure how to update this:

https://github.com/Qiskit/qiskit-metapackage/blob/2d30932b500a5b8d28fa823f9474df902b0afd08/docs/index.rst?plain=1#L31-35

Sounds like it should be https://qiskit.org/providers, right? I can make that change in my PR.

README.md
```diff
@@ -70,10 +70,9 @@ install them manually, as an optional extra
 ### Using quantum services
 
 If you are interested in using quantum services (i.e. using a real quantum
-computer, not a simulator) you can look at the Qiskit Partners program for
-partner organizations that have provider packages available for their offerings:
+computer, not a simulator) you can check https://qiskit.org/providers/#hardware
+that have provider packages available for their offerings.
```
HuangJunye: I would suggest to remove the emphasize of real hardware and just use services as a general term. And then just refer to the providers page instead of just the hardware section. Because there are multiple sections of the provider page contain services with access to hardware: 1) native primitives, 2) quantum hardware, 3) multi-platforms

```
If you are interested in using quantum services, you can check [the Providers page](https://qiskit.org/providers/) for the list of available providers that Qiskit supports.
```
(I can't make github suggestion directly because the first line was not changed by the PR)

HuangJunye (commented):
I added a suggestion.

README.md
javabster: I like Junye's phrasing, but i think it sounds better to say "you can check out.." not just "you can check..". If you just use the word "check" it implies you are checking or validating that something is correct, whereas the phrase "check out" implies you are looking into something out of interest (which I think is more aligned with the tone you're trying to use here?)

jakelishman (dismissed):
Seems fine to me, and aside from old release notes and bits and bobs around the old release process that don't need changing, there doesn't appear to be any other references to partners in the repo.

I don't know if you want to mention that qiskit.org/providers will always just be a partial list (since there's no special access needed - you just publish a package).

Since we're modifying things in a readme for a repo that's about to be archived, is there a plan for how this will migrate to Terra?  As a second aside, I also notice that all the build-status/licence/etc badges in the README still use `Qiskit/qiskit` rather than `Qiskit/qiskit-metapackage`.

README.md
```diff
@@ -1,5 +1,7 @@
 ![Image](https://raw.githubusercontent.com/Qiskit/qiskit/master/images/qiskit_header.png)
 
+
+
```
Eric-Arellano: Uninentional?

Eric-Arellano (dismissed):

README.md
1ucian0: yeap.. fixing in 3a3469db08c7d25b50f23370f3aa50109fac7aa9

1ucian0:
> Since we're modifying things in a readme for a repo that's about to be archived, is there a plan for how this will migrate to Terra?

You mean the README file? I can follow up with a warning box saying "this repo is going to be archived soon".

> As a second aside, I also notice that all the build-status/licence/etc badges in the README still use `Qiskit/qiskit` rather than `Qiskit/qiskit-metapackage`.

done in c166bb48cd8bad1a38bf398240be2b9518dbdafb

Eric-Arellano (approved):

PR #1753, opened 2023-06-02
Make changes to bump version

jlapeyre:
Bump version to 0.43.11

PR #1754, opened 2023-06-05
Add/fix additions to release notes for recent release 0.43.1

jlapeyre:
The release notes for the most recent release did not follow the required template. This PR provides a fix.

Release 0.43.1 was made for this package because Terra 0.24.1 was released. See PR #1753.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -28,7 +28,7 @@ Qiskit 0.43.1
 .. _Release Notes_Terra_0.24.1:
 
 Terra 0.24.1
-======
+============
 
 .. _Release Notes_Terra_0.24.1_Prelude:
 
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -27,7 +27,7 @@ Qiskit 0.43.1
 
 .. _Release Notes_Terra_0.24.1:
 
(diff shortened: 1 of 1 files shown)


Eric-Arellano (dismissed):

docs/release_notes.rst
```diff
@@ -27,7 +27,7 @@ Qiskit 0.43.1
 
 .. _Release Notes_Terra_0.24.1:
 
-0.24.1
+Terra 0.24.1
 ======
```
jakelishman: The underline needs expanding - it's an rST warning to have a too-short underline (but I think it's permissible to be too long - it's kind of weird).

jlapeyre: I don't see why it should not be required to be exactly the same length as the text.

jakelishman: No me neither, but weirdly [the spec explicitly says they can be longer](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections).

jakelishman (approved):
Thanks for the fix.  Fortunately, we're in the last throes of needing to screw around with this document - once the metapackage is properly retired, we'll be able to use the more standard `reno` workflows from within Terra.

PR #1755, opened 2023-06-06
Remove extras installers for apps modules and experiments.

blakejohnson:
Code change to accompany #1751. Removes apps module extra installers from `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,29 +30,6 @@ requirements = [
 ]
 
 
-optimization_extra = [
-    "qiskit-optimization>=0.4.0",
-]
-
-
-finance_extra = [
-    "qiskit-finance>=0.3.3",
-]
-
-
-machine_learning_extra = [
-    "qiskit-machine-learning>=0.4.0",
-]
-
-
-nature_extra = [
-    "qiskit-nature>=0.4.1",
-]
-
-experiments_extra = [
-    "qiskit-experiments>=0.2.0",
-]
-
 visualization_extra = [
     "matplotlib>=2.1",
     "ipywidgets>=7.3.0",
@@ -101,16 +78,6 @@ setup(
     python_requires=">=3.7",
     extras_require={
         "visualization": visualization_extra,
-        "all": optimization_extra
-        + finance_extra
-        + machine_learning_extra
-        + nature_extra
-        + experiments_extra
-        + visualization_extra,
-        "experiments": experiments_extra,
-        "optimization": optimization_extra,
-        "finance": finance_extra,
-        "machine-learning": machine_learning_extra,
-        "nature": nature_extra,
+        "all": visualization_extra,
     },
 )


Eric-Arellano (changes requested):
This is disruptive for end users because it risks breaking their build pipelines, such as CI and packaging, in two ways:

1. If they explicitly set the extra, their installation will now error that the extra is not found.
2. If they were using `qiskit[all]`, and also using any of those extras, they will now be missing those dependencies so will fail at import time when using things like Qiskit Nature.

I agree with the direction of this PR. I'm only blocking until we have agreed on the plan for how to minimize user disruption. Here are some suggestions:

1. This should be accompanied with a version bump of qiskit-metapackage. That reduces the fallout—if people pin to minor versions, they won't be broken by this.
2. How are we communicating this change to the community? Such as Slack or user groups?
3. We should get rid of `[all]` to address problem 2. That problem is particularly concerning to me because it is confusing/silent; pip won't give an explicit error message saying "this extra doesn't exist!" It instead will be a cryptic failure at import time, and users not familiar with this change or Python packaging will be confused.

Unfortunately, I don't think there's a way for us to issue runtime deprecation warnings. In setup.py, we don't know what extras the user has selected. The only option is to add runtime warnings to the extras themselves, but that's not feasible because we don't know how the library got installed, e.g. directly vs the metapackage.

blakejohnson:
We have to accept this disruption because these other packages (with the exception of `qiskit-experiments`) are no longer receiving active development by the core Qiskit team. So, either you let people continue to install dependencies that we don't maintain, or you force them to realize that those packages aren't included in Qiskit and consciously choose to install them anyway.

Eric-Arellano:
> We have to accept this disruption 

Sure, there will be disruption no matter what. But it's not binary. There are meaningful ways to reduce the disruption to users. That's valuable to keep trust in Qiskit; one of the quickest ways to burn trust is to break people's CI.

So, I am curious what you think about how we can reduce this damage of trust? What are your thoughts on the above 3 suggestions?

mtreinish:
The working plan for doing this migration was to do it after the next big release at the end of July. That release will be marking the final "metapackage release", and as part of that large upgrade note about all the changes being made (eol for ibmq, aer no longer being included, etc) we're planning to document that release, qiskit 0.44.0, is the final release that has these extras included and document how users will need to update their potential usage of them. I think this should wait until after that release and continue on that path.

blakejohnson:
Sounds good, @mtreinish

javabster:
@Eric-Arellano fwiw we've been slowly trying to move people away from installing the application modules via the metapackage for a little while now (see the metapackage section of [this post](https://medium.com/qiskit/what-is-qiskit-your-guide-to-navigating-the-qiskit-cosmos-in-2022-7cc415426f31) from a while back). And when the changes in this PR go into effect with the next big release we'll make sure to warn users in our usual blog/youtube/twitter comms

1ucian0:
> The working plan for doing this migration was to do it after the next big release at the end of July. 

See https://github.com/Qiskit/RFCs/issues/41

jakelishman:
Just for fun: _technically_ we can emit some degree of runtime warning if we really wanted to.  The trick is to make a dummy package that we upload to PyPI and add to the given extras with a no-collision name like `qiskit_deprecation_check_nature` or whatever.  Then there's something that can be tested for at runtime (or even install time, if we're being particularly bad citizens of Python, since we only provide an sdist for the metapackage) using `importlib.metadata`, and it wouldn't be installed by any standard means.

Eric-Arellano:
Given the updated plan in https://github.com/Qiskit/RFCs/issues/41, I'm going to close this so that we keep the PR list clean with only currently active PRs. We can reopen this PR when ready.

> In July release, mention that optionals are going to be removed in October release
> In October release, remove optionals from metapackage. Make the terra and metapackage optionals, the same.

1ucian0:
> Just for fun: _technically_ we can emit some degree of runtime warning if we really wanted to.

I'm not sure if it is worthy.

While it is hard to know how used are extras, there is no much evidence of the use of _these_ extras. There are [6 questions in stackoverflow that mention qiskit installation](https://stackoverflow.com/search?q=%22pip+install+qiskit%22): one of them [mentioning `visualization`](https://stackoverflow.com/questions/69802802/install-ibm-qiskit-for-quantum-computer-simulation), nothing about the rest.  Similarly, in stackexchange, 12 questions with `pip install qiskit`, none of them using extras. The only entry I found in google (besides installation instruction) is [this one](https://colab.research.google.com/drive/1OLAfeHPFhdasNxci0gtmrovGsYmKz64b?usp=sharing) for `nature`, [these](https://gitlab.qpiai.tech/qpiquantum/optimisation/qpiai-opt-api) [two](https://entangledquery.com/t/how-to-use-qiskit-in-jupyter-notebook/5/) for `optimization`, [one](https://www.chegg.com/homework-help/questions-and-answers/1-starting-iris-example-given--experiment-various-architectural-choices-least-2-per-item-d-q112952688) for `machine-learning`, and [this one](https://pyqml.substack.com/p/quantum-normal-distribution) for `finance`. 

I think the reason is because there is no advantage wrt to just installing the package itself. `all` seems slightly more used, by some [tutorials](https://community.ibm.com/community/user/ibmz-and-linuxone/blogs/jack-woehr1/2022/05/20/install-ibm-qiskit-on-linuxone-ubuntu) [and](https://www.kaggle.com/code/srivnaman/quantum-machine-learning) [blogpost](https://blueqat.com/derwind/df279f94-2020-4eb0-b694-b6797e62de47) out there, but now they have their own namespace, I think is fair to assume they need their own packages.

PR #1757, opened 2023-06-09
Add back tweedledum for Grover tutorial

Eric-Arellano:
https://qiskit.org/documentation/tutorials/algorithms/07_grover_examples.html is currently missing its output.

We set `QISKIT_DOCS_BUILD_TUTORIALS` in our docs publishing. So it should be sufficient to only update this repository to fix the problem.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,6 +13,9 @@ pillow>=4.2.1
 seaborn>=0.9.0
 nbsphinx
 cvxpy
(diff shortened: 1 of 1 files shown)


Eric-Arellano:
As discussed in this week's dev meeting, I am discussing with Luciano what the long term plan is for Tweedledum and if it makes sense to write a minimal Rust library that replicates the functionality we need.

mtreinish (approved):

PR #1760, opened 2023-06-12
Update 'Other API' links on docs homepage.

blakejohnson:
Reference the non-deprecated IBM provider and add the Qiskit Runtime provider.


Eric-Arellano (commented):
My understanding is that we've been planning on removing these from the left sidebar entirely. Can you please confirm @javabster @HuangJunye? And that we needed to wait more time to remove IBMQ Provider.

1ucian0:
These links refer to the packages included in the metapackage. All these links will be removed in July release.

PR #1768, opened 2023-07-03
[experiment] Try no page per method

Eric-Arellano:
I want to see how much faster docs builds are using the Pytorch theme.


Eric-Arellano:
Replaced by https://github.com/Qiskit/qiskit-terra/pull/10455.

PR #1769, opened 2023-07-07
Add intersphinx extension for linking to other orgs

psschwei:
### Summary

Adding intersphinx extension to allow for linking to other orgs in docs

### Details and comments

I added links for `matplotlib.figure.Figure` return types to the visualization docs in https://github.com/Qiskit/qiskit-terra/pull/10389 , and @Eric-Arellano  requested in https://github.com/Qiskit/qiskit-terra/pull/10389#pullrequestreview-1518994619 to make that change here as well. 

It didn't appear that the intersphinx extension was being used in this repo (I didn't find it in `docs/conf.py`), so I added the extension and all the mappings from qiskit-terra. Let me know if that's not the correct approach to take here.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,14 +84,6 @@ nbsphinx_prolog = """
 
 """
 
-intersphinx_mapping = {
-    "rustworkx": ("https://qiskit.org/ecosystem/rustworkx/", None),
-    "qiskit-ibm-runtime": ("https://qiskit.org/ecosystem/ibm-runtime/", None),
-    "qiskit-aer": ("https://qiskit.org/ecosystem/aer/", None),
-    "numpy": ("https://numpy.org/doc/stable/", None),
-    "matplotlib": ("https://matplotlib.org/stable/", None),
-}
-
 panels_css_variables = {
     "tabs-color-label-active": "rgb(138, 63, 252)",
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -83,6 +84,14 @@
 
 """
 
+intersphinx_mapping = {
+    "rustworkx": ("https://qiskit.org/ecosystem/rustworkx/", None),
+    "qiskit-ibm-runtime": ("https://qiskit.org/ecosystem/ibm-runtime/", None),
+    "qiskit-aer": ("https://qiskit.org/ecosystem/aer/", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),
+    "matplotlib": ("https://matplotlib.org/stable/", None),
+}
```
Eric-Arellano: Nit, can you please move this down to be in the section with `extlinks`?

Eric-Arellano (dismissed):
Oh wow, good catch that intersphinx hasn't been used...this is exactly the type of complexity we suffer from with docs being split between the two repositories. Thank you!

docs/conf.py
psschwei: done (sorry, didn't see that section)

Eric-Arellano (approved):

PR #1773, opened 2023-07-11
Fix Black 23.7.0 issue

Eric-Arellano:
`extend-exclude` should have been a string. We simply use the old behavior of Black, which was to call `repr()` on the list. 

This also properly pins Black's version. I didn't realize they're using cal-ver, so `~=` doesn't work properly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [tool.black]
 line-length = 100
 target-version = ["py37", "py38", "py39", "py310", "py311"]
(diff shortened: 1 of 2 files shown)


mtreinish (approved):
I think the `~=` should work for a calver because we should just use the latest in the 23.x series which will be compatible per https://black.readthedocs.io/en/stable/the_black_code_style/index.html#labels-stability-policy but a hard pin is fine with me too. AFAIK we've never encountered a bug in black that required a bugfix so using a fixed version is fine.

PR #1776, opened 2023-07-17
Remove Aer tutorials

Eric-Arellano:
These now live in Aer itself, per https://github.com/Qiskit/qiskit-aer/pull/1768. Terra docs should no longer include Aer directly.

Once this PR lands, we will delete the tutorials from the qiskit-tutorials repo.


javabster (approved):
LGTM 🚀

sg495:
This makes sense, but it would have been better to include a redirect on the affected Terra documentation pages.

Eric-Arellano:
I agree with you @sg495. Pardon my oversight in not doing that earlier, and thanks for advocating for that! I opened up a PR in the config for qiskit.org to set up redirects from the `tutorials/simulators` folder to go to https://qiskit.org/ecosystem/aer/tutorials/index.html.

Eric-Arellano:
The redirects are now live. Thanks for the feedback @sg495!

PR #1789, opened 2023-08-04
Simplify README

mtreinish:
### Summary

After the Qiskit 0.44.0 release the qiskit-metapackage is only installing qiskit-terra now. This commit updates the readme to reflect this. Previously it was still referring to the other elements being installed in the metapackage.

### Details and comments


jakelishman (approved):
Best to get this updated here now, and then we can update Terra's readme properly once everything's joined up.

